Issue #87, opened 2015-04-15
Add support for function currying


LordBlackhole:
**-Enhancment**

Something I've been finding myself doing quite a bit recently is writing extra lambdas for currying my functions. Turning Functions into Suppliers and BiFunctions into Functions by partially applying the first argument. I was writing some static helper methods for simplifying this when I realized that it would be a lot easier and make more sense to just have those methods on the Function interfaces themselves. 

So for example, Function2 would have ....

```java
default Function1<T2, R> curry(T1 v1) {
    return v2 -> apply(v1, v2);
}

default Function0<R> curry(T1 v1, T2 v2){
    return () -> apply(v1, v2)
}
```

While not *technically* the proper definition of Currying, which would require having only one method returning `Function<T1, Function<T2, R>>`, I feel like this would produce code that would be easier to read and more familiar to a Java developer. For example if you had a Function4 and wanted to apply the first three arguments, it'd be a simple matter of `func.curry(var1, var2, var3)` instead of `func.curry().apply(var1).apply(var2).apply(var3)`

I'm really liking this library and I would be happy to write up the boiler-plate for this and make a pull request. I just wanted to know if there would be any interest in it.

cdshines:
I've got a couple of places where this would fit, so +1.

lukaseder:
Sounds reasonable!

danieldietrich:
Hi!

I'm about to steal that idea for Javaslang, naming it 'partial function application':

```java
// pseudo / no valid code
Function1<T2, R> Function2<T1, T2, R>.apply(T1 t1) = (T2 t2) -> apply(t1, t2);
```

lukaseder:
@danieldietrich Stealing the idea and a couple of users at the same time? Sneaky! ;)

danieldietrich:
Whooops...

A blog post would steal the rest of the users ;-)

lukaseder:
Don't worry. I'd still love to publish that blog post. jOOL is not a full-fledged functional API (we don't have time for that). We just fill the gaps where very obvious things are missing from the JDK.

danieldietrich:
You will get it! I'm currently collecting ideas...

LordBlackhole:
@danieldietrich  You know, you don't actually have to steal my idea. I can fork javaslang and write it up there too.

danieldietrich:
@LordBlackhole That would be great but it is already finished. I mentioned you [here](https://github.com/javaslang/javaslang/issues/153).

[...]

Issue #54, opened 2014-09-28
Add Functions.not(Predicate), and(Predicate...), or(Predicate...)


ajermakovics:
Hello

Since there is no `not(Predicate)` function in Java 8:
http://stackoverflow.com/q/21488056/1180621

seems like it would be a useful addition to jOOL.

PR #88, opened 2015-04-16
Added Curry methods to Functions

LordBlackhole:
As per: #87 

Added curry() methods to all Functions, allowing them to be partially applied and return a new function that accepts the remaining arguments.

### DIFF
--- a/src/main/java/org/jooq/lambda/function/Function1.java
+++ b/src/main/java/org/jooq/lambda/function/Function1.java
@@ -55,7 +55,7 @@ public interface Function1<T1, R> extends Function<T1, R> {
     }
 
     /**
-     * Partially apply this function to return a {@link org.jooq.lambda.function.Function0}
+     * Partially apply this function to the arguments.
      */
     default Function0<R> curry(T1 v1){
         return () -> apply(v1);
--- a/src/main/java/org/jooq/lambda/function/Function1.java
+++ b/src/main/java/org/jooq/lambda/function/Function1.java
@@ -16,6 +16,7 @@
 package org.jooq.lambda.function;
 
 import java.util.function.Function;
+
 import org.jooq.lambda.tuple.Tuple1;
 
 /**
@@ -51,5 +52,12 @@ public interface Function1<T1, R> extends Function<T1, R> {
      */
     static <T1, R> Function1<T1, R> from(Function<T1, R> function) {
         return function::apply;
+    }
+
+    /**
+     * Partially apply this function to return a {@link org.jooq.lambda.function.Function0}
+     */
+    default Function0<R> curry(T1 v1){
+        return () -> apply(v1);
     }
 }
--- a/src/main/java/org/jooq/lambda/function/Function2.java
+++ b/src/main/java/org/jooq/lambda/function/Function2.java
@@ -54,10 +54,16 @@ public interface Function2<T1, T2, R> extends BiFunction<T1, T2, R> {
         return function::apply;
     }
 
+    /**
+     * Partially apply this function to the arguments.
+     */
     default Function1<T2, R> curry(T1 v1){
         return v2 -> apply(v1, v2);
     }
 
+    /**
+     * Partially apply this function to the arguments.
+     */
     default Function0<R> curry(T1 v1, T2 v2){
         return () -> apply(v1, v2);
     }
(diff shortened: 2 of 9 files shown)


lukaseder:
Nice, thank you. Are you contributing this under the terms of the [Apache License 2.0](http://www.apache.org/licenses/LICENSE-2.0)?

danieldietrich:
'Morning guys,
I'm not a licensing expert but by contributing to a project which is under Apachy 2.0 license, this should be clear, isn't it?

lukaseder:
Says who? Some random lawyer at Red Hat? :-)

danieldietrich:
Hehe, don't know :-) Read this https://www.apache.org/licenses/LICENSE-2.0 and interpreted def. of Contribution + Grant of * (2. and 3.)...

lukaseder:
Eh, actually, it's here:

> 5. Submission of Contributions. Unless You explicitly state otherwise, any Contribution intentionally submitted for inclusion in the Work by You to the Licensor shall be under the terms and conditions of this License, without any additional terms or conditions. Notwithstanding the above, nothing herein shall supersede or modify the terms of any separate license agreement you may have executed with Licensor regarding such Contributions.

Interesting, I haven't noticed this, so far. Thanks!

danieldietrich:
Interesting is the last sentence

> Notwithstanding the above, nothing herein shall supersede or modify the terms of any separate license agreement you may have executed with Licensor regarding such Contributions.

Does this has an impact to a dual license? I think not because we speak about contribution and not usage.

lukaseder:
> Does this has an impact to a dual license?

What kind of impact were you thinking of?

danieldietrich:
No, this was a 'Denkfehler'.

LordBlackhole:
@lukaseder Oh yes, I wrote that lift method in a spur-of the moment to get the method into a function, and then immediately realized how useful it could be everywhere else. 

I saw how you mentioned in #54 having a single point of entry for the DSL. Is there a place for that yet? How about a "Functions" class next to Function0-8? Could be a dump-all for all Function-related static helper methods. 

The thing about lift() is that it works great unless the method you want to lift is overloaded. Then you must provide the generics, which is tedious. 

Another thing it can't do is pick up a method with a return type of void, as those aren't functions. Which got me thinking about Consumers, and I realized there is nothing preventing making Function1 and Function2 extend Consumer and BiConsumer, they'd just throw away their results. 

Might not be that terribly useful, but I've been using the Stream.peak() for logging to help debug them, and because those methods return void you can't use them with any of the Functions without wrapping them manually. 

Oh, and yes I respect the license, I thought that was implied.

lukaseder:
>  saw how you mentioned in #54 having a single point of entry for the DSL. Is there a place for that yet?

No.

> How about a "Functions" class next to Function0-8? Could be a dump-all for all Function-related static helper methods.

I know that many people like to group "global functions". I prefer to have only a single class where all static methods / "global functions" go, so people can just static-import `com.example.SingleClass.*` to have all the awesome tools on the classpath. It sucks when you have to find a method in dozens of utility classes.

> Another thing it can't do is pick up a method with a return type of void, as those aren't functions. Which got me thinking about Consumers, and I realized there is nothing preventing making Function1 and Function2 extend Consumer and BiConsumer, they'd just throw away their results.

Could you show an example? I personally don't think it's a good idea to have those types extend each other... That seems to be calling for trouble later on.

> Oh, and yes I respect the license, I thought that was implied.

Yes, it is. I stood corrected.

LordBlackhole:
All I was thinking was 

```java
public class Function1<T1, R> extends Function<T1, R>, Consumer<T1> {

    @Override
    default accept(T1 v1) {
        apply(v1);
    }

...
}
```

Although I do see what you mean. Some way of lifting a void method into a function so we  can use the curry and other methods might be useful though. Now I truly understand why Scala uses Unit instead of void. Another option would be to have Consumer0 - 8, but I think that's far too much effort. I don't know, it might not be that necessary. Just having lift() methods that take Consumers and turn them info Functions would work too, though there isn't an obvious return type.

lukaseder:
I see, yes that's what I thought you meant. But I have a very uneasy feeling with this. I'm sure this will cause compile-time ambiguities somewhere down the line.

> Now I truly understand why Scala uses Unit instead of void.

Yes. `void` is such a horrible wart in the Java type system.

> Another option would be to have Consumer0 - 8, but I think that's far too much effort

No, I think that would certainly make sense, eventually...

> Just having lift() methods that take Consumers and turn them info Functions would work too

Yep. In the meantime, this issue seems to have been resolved: http://stackoverflow.com/q/23430854/521799, so you're safe to do that.

PR #90, opened 2015-05-10
Added JOOL utility class with lift methods.

DillonJettCallis:
As discussed in #88 I've added a class with static "lift" methods. 

@lukaseder - You said you wanted only a single class but didn't say what it should be called or anything so I hope that my choice of "JOOL" is fine. 

The lift[n] methods are to help lifting overridden methods. As long as
the overrides have unique numbers of arguments using the right lift[n]
should work. If they aren't unique then the user has no choice but to
use the full generics.

I wasn't really sure what the Javadoc should look like. JOOL Test class
isn't exhaustive either, but should demonstrate the principal.

### DIFF
--- a/src/main/java/org/jooq/lambda/JOOL.java
+++ b/src/main/java/org/jooq/lambda/JOOL.java
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) 2014-2015, Data Geekery GmbH, contact@datageekery.com
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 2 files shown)


tkroman:
This seems peculiar and interesting. I'd like tests to show more non-synthetic use-cases (or you might just include them into commit's message, please?)

LordBlackhole:
The primary reason I have is so I can lift a method and then curry it, but any other Function methods could be a reason as well. 

I don't really have a great example off the top of my head, other than logging (where I would like to lift the slf4j debug method and then curry the string message so the object can be applied), but that isn't possible until there are Consumer[n] with those methods since debug returns null. 

Here is the best example I can come up with, granted that it's till synthetic. 
```java
public void doStuff(){
   List<SomeType> list = getListOfSomething();
   StreamContext someContextObject = getContext();

   list.stream().map(lift(this::convert).curry(someContextObject)).forEach(e -> /* do actual stuff */);
}

private SomeOtherType convert(StreamContext context, SomeType thing){
   if(context.getValueA() < 0)
      return new SomeNegativeType(thing);
   else
      return new SomePositiveType(thing);
}
```

lukaseder:
Thanks for the contribution and sorry for the delay.

I'm not sure if `lift()` would be the appropriate wording here, given that it is usually used for lifting a partial function into a total function. In your example, `this::convert` is already a total function, isn't it? How about just calling it `function()`, `func()` or `f()` (or all of them)?

lukaseder:
I'm reviewing this, and I don't think that this will be very useful or expected to users...

PR #97, opened 2015-07-20
[#93] Add Seq.exists(Predicate)

### LINKED ISSUE #93: Add Seq.exists(Predicate)
Currently I'm using filter(Predicate).findFirst().isPresent() and .exists(Predicate) would be a little shorter.

stephenh:
Fixes #93.


lukaseder:
Thanks for the suggestion. Won't pull this one. Rationale here: https://github.com/jOOQ/jOOL/issues/93#issuecomment-122772987

PR #116, opened 2015-09-08
Add Maven dependency to README.md

tkroman:
Since jOOL is all about simplifying the life, it's natural that plugging it into users' projects should be as easy as using it.


lukaseder:
True, why not. Thanks for the contribution!

PR #121, opened 2015-09-15
Fix minor javadoc copy/paste error

netzwerg:


lukaseder:
Great catch, thank you very much!

PR #131, opened 2015-10-29
Added javadocs for Seq methods regarding parallel and unordered Streams

### LINKED ISSUE #130: Some methods on Seq are not correctly implemented
I was debugging some parallel streams code that uses `Seq` and found this code inside `Seq`.

```java
    // These methods have no effect
    // ----------------------------

    @Override
    default Seq<T> sequential() {
        return this;
    }

    @Override
    default Seq<T> parallel() {
        return this;
    }
```

Well, that's... unexpected. That also applies to some other methods like `unordered()`.

So may we consider a _bug_ not documenting something THAT important. IMHO. Javadocs at least.

Also `SeqImpl` has this code:

```java
    @Override
    public boolean isParallel() {
        return false;
    }
```

So there is clearly a design decision. By the way, @lukaseder, could you please explain why?

sargue:
Fixes #130.
As discussed in the issue we're hinting on the javadocs that some Stream
methods just are no-ops on a sequential and ordered stream.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -5235,16 +5235,34 @@ public interface Seq<T> extends Stream<T>, Iterable<T> {
     // These methods have no effect
     // ----------------------------
 
+    /**
+     * Returns this stream. All Seq streams are sequential, hence the name.
+     *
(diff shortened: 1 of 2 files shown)


lukaseder:
Perfect, thank you very much!

Issue #134, opened 2015-11-03
Implementing InnerJoin with HashMap


ksranga:
Hi - do you see a performance benefit in implementing the InnerJoin by doing a lookup against HashMap after converting one of the List to a Map using the join column as the key?

lukaseder:
Indeed, implementing a hash join rather than a nested loop might speed things up in certain situations. Although, care must be taken in the event of duplicates. It's perfectly fine for an inner join to operate on duplicates, e.g. the following test must pass:

```java
        assertEquals(asList(
            tuple(1, 1),
            tuple(1, 2),
            tuple(1, 1),
            tuple(1, 2)),
            Seq.of(1, 1).innerJoin(Seq.of(1, 2), (a, b) -> true).toList());
```

Would you be interested in implementing this improvement?

ksranga:
Yes - I can help out. To handle duplicates, the Map we have to build would have a List as its value

Map<TupleN, List<T>>

TupleN are the columns in the join condition.

lukaseder:
Indeed, this would work. I suspect that we should implement this optimisation only after a given threshold, e.g. when the argument list size exceeds 8 or so, as the map creation and hashcode / equals overhead might exceed the benefit of using a map

ksranga:
I am just wrapped up this feature. How do I go about getting access so I can create a pull request?

lukaseder:
Just create a pull request... In order to do so, commit your changes to your own fork, and click on the pull request button on GitHub

lukaseder:
I'm closing this as "won't fix". While with sufficent knowledge about the nature of the joined sets that participate in a join operation (as in SQL, when indexes and constraints are known), a nested loop might just be good enough in a more general use-case.

The suggestions exposed in this discussion are not doable, as they

1. Change the semantics of a relational join operation in our API
2. Change the API itself

PR #139, opened 2015-11-17
innerJoin implementation with Map issue 134

ksranga:
implementation of innerJoin using Map lookup

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -68,20 +68,7 @@ import java.util.stream.StreamSupport;
 
 import javax.annotation.Generated;
 
-import org.jooq.lambda.function.Function10;
-import org.jooq.lambda.function.Function11;
-import org.jooq.lambda.function.Function12;
-import org.jooq.lambda.function.Function13;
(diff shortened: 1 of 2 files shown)


lukaseder:
Thank you very much for this suggestion. I will comment on individual lines of code, directly here on GitHub

lukaseder:
I understand the complexity discussion, but I don't agree with the (public!) API from two points of view:

- It is rather irregular. The `Map` type is not a first-class citizen in jOOλ. So, this method is rather unexpected
- I would have preferred to rewrite the existing functionality rather than to introduce new functionality

I'm sorry for the misunderstanding this may have caused. But I really don't agree with the suggested API

lukaseder:
Also, I've just noticed, your suggested implementation is not implementing the semantics of "inner join". A join operation forms a cartesian product (`flatMap()` with Java 8), filtering out tuples using the join predicate. Your suggested implementatation is not forming any cartesian products.

I'm closing this pull request for now. We'll think of something to implement #134. But it really isn't a pressing issue. For many situations a nested loop is an OK implementation for inner joins. Hash joins aren't really easy to implement, especially if we don't have any meta information about the joined collections (sizes, cardinalities, indexes, etc.)

PR #140, opened 2015-11-29
Changed readme description of Seq to better describe it

sargue:
Seq is short of Sequential and the original text doesn't warn about that
and it even talks about "parallellisation" which can lead to
misunderstandings.


lukaseder:
I agree, thanks for the suggestion!

PR #172, opened 2016-01-08
Seq2 added - it is like Seq<Tuple2> but with BiFunction, BiPredicate, etc.

korgov:
Seq2 added - it is like Seq<Tuple2> but with BiFunction, BiPredicate, etc. in method args for more clean code.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq2.java
+++ b/src/main/java/org/jooq/lambda/Seq2.java
@@ -1,0 +1,274 @@
+package org.jooq.lambda;
+
+import org.jooq.lambda.function.Function3;
+import org.jooq.lambda.tuple.Tuple;
+import org.jooq.lambda.tuple.Tuple2;
+
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Spliterator;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.ToDoubleBiFunction;
+import java.util.function.ToIntBiFunction;
+import java.util.function.ToLongBiFunction;
+import java.util.stream.DoubleStream;
+import java.util.stream.IntStream;
+import java.util.stream.LongStream;
+import java.util.stream.Stream;
+
+/**
+ * @author Kirill Korgov (kirill@korgov.ru)
+ *         30.09.2015 19:21
+ */
+public interface Seq2<T1, T2> extends Seq<Tuple2<T1, T2>> {
+
+    /**
+     * Wrap an <code>Iterable</code> into a <code>Seq2</code>.
+     */
+    static <T1, T2> Seq2<T1, T2> seq(final Iterable<Tuple2<T1, T2>> iterable) {
+        return seq(Seq.seq(iterable));
+    }
+
+    /**
+     * Wrap an <code>Iterator</code> into a <code>Seq2</code>.
+     */
+    static <T1, T2> Seq2<T1, T2> seq(final Iterator<Tuple2<T1, T2>> iterator) {
+        return seq(Seq.seq(iterator));
+    }
+
+    /**
+     * Wrap a <code>Spliterator</code> into a <code>Seq2</code>.
+     */
+    static <T1, T2> Seq2<T1, T2> seq(final Spliterator<Tuple2<T1, T2>> spliterator) {
+        return seq(Seq.seq(spliterator));
+    }
+
+    /**
+     * Wrap a <code>Map</code> into a <code>Seq2</code>.
+     */
+    static <K, V> Seq2<K, V> seq(final Map<K, V> map) {
+        return Seq2.seq(Seq.seq(map));
+    }
+
+    /**
+     * Wrap an <code>Optional</code> into a <code>Seq2</code>.
+     */
+    static <T1, T2> Seq2<T1, T2> seq(final Optional<Tuple2<T1, T2>> optional) {
+        return seq(Seq.seq(optional));
+    }
+
+    static <T1, T2> Seq2<T1, T2> seq(final Stream<Tuple2<T1, T2>> stream) {
+        return seq(Seq.seq(stream));
+    }
+
+    static <T1, T2> Seq2<T1, T2> seq(final Seq<Tuple2<T1, T2>> src) {
+        if (src instanceof Seq2) {
+            return (Seq2<T1, T2>) src;
+        }
+        return new Seq2Impl<>(src);
+    }
+
+    static <T1, T2> Seq2<T1, T2> zip(final Seq<T1> s1, final Seq<T2> s2) {
+        return seq(s1.zip(s2));
+    }
+
+    static <T1, T2> Seq2<T1, T2> zip(final Stream<T1> s1, final Stream<T2> s2) {
+        return zip(Seq.seq(s1), Seq.seq(s2));
+    }
+
+    static <T1, T2> Seq2<T1, T2> zip(final Iterable<T1> s1, final Iterable<T2> s2) {
+        return zip(Seq.seq(s1), Seq.seq(s2));
+    }
+
+    static <T1, T2> Seq2<T1, T2> zip(final Iterator<T1> s1, final Iterator<T2> s2) {
+        return zip(Seq.seq(s1), Seq.seq(s2));
+    }
+
+    static <T> Seq2<T, Long> zipWithIndex(final Seq<T> stream) {
+        return Seq2.seq(Seq.zipWithIndex(stream));
+    }
+
+    static <T> Seq2<T, Long> zipWithIndex(final Iterable<T> iterable) {
+        return Seq2.zipWithIndex(Seq.seq(iterable));
+    }
+
+    static <T> Seq2<T, Long> zipWithIndex(final Stream<T> stream) {
+        return Seq2.zipWithIndex(Seq.seq(stream));
+    }
+
+    default <U> U foldLeft(final U seed, final Function3<U, ? super T1, ? super T2, U> function) {
+        return Seq.super.foldLeft(seed, (u, v) -> function.apply(u, v.v1, v.v2));
+    }
+
+    default Seq2<T1, T2> filter(final BiPredicate<? super T1, ? super T2> predicate) {
+        return Seq2.seq(filter(t -> predicate.test(t.v1, t.v2)));
+    }
+
+    default Seq2<T1, T2> filter1(final Predicate<? super T1> predicate) {
+        return Seq2.seq(filter(t -> predicate.test(t.v1)));
+    }
+
+    default Seq2<T1, T2> filter2(final Predicate<? super T2> predicate) {
+        return Seq2.seq(filter(t -> predicate.test(t.v2)));
+    }
+
+    default <R> Seq<R> map(final BiFunction<? super T1, ? super T2, R> fu2) {
+        return map(t -> fu2.apply(t.v1, t.v2));
+    }
+
+    default <R1> Seq2<R1, T2> map1(final BiFunction<? super T1, ? super T2, R1> fu2) {
+        return mapToT2((v1, v2) -> Tuple.tuple(fu2.apply(v1, v2), v2));
+    }
+
+    default <R2> Seq2<T1, R2> map2(final BiFunction<? super T1, ? super T2, R2> fu2) {
+        return mapToT2((v1, v2) -> Tuple.tuple(v1, fu2.apply(v1, v2)));
+    }
+
+    default <R1> Seq2<R1, T2> map1(final Function<? super T1, R1> fu2) {
+        return map1((v1, v2) -> fu2.apply(v1));
+    }
+
+    default <R2> Seq2<T1, R2> map2(final Function<? super T2, R2> fu2) {
+        return map2((v1, v2) -> fu2.apply(v2));
+    }
+
+    default <R1, R2, R extends Tuple2<R1, R2>> Seq2<R1, R2> mapToT2(final BiFunction<? super T1, ? super T2, R> fu2) {
+        return Seq2.seq(map(t -> fu2.apply(t.v1, t.v2)));
+    }
+
+    default Seq<T1> keys() {
+        return v1s();
+    }
+
+    default Seq<T2> values() {
+        return v2s();
+    }
+
+    default Seq<T1> v1s() {
+        return map(Tuple2::v1);
+    }
+
+    default Seq<T2> v2s() {
+        return map(Tuple2::v2);
+    }
+
+    default IntStream mapToInt(final ToIntBiFunction<? super T1, ? super T2> mapper) {
+        return mapToInt(t -> mapper.applyAsInt(t.v1, t.v2));
+    }
+
+    default LongStream mapToLong(final ToLongBiFunction<? super T1, ? super T2> mapper) {
+        return mapToLong(t -> mapper.applyAsLong(t.v1, t.v2));
+    }
+
+    default DoubleStream mapToDouble(final ToDoubleBiFunction<? super T1, ? super T2> mapper) {
+        return mapToDouble(t -> mapper.applyAsDouble(t.v1, t.v2));
+    }
+
+    default <R> Seq<R> flatMap(final BiFunction<? super T1, ? super T2, ? extends Stream<? extends R>> fu2) {
+        return flatMap(t -> fu2.apply(t.v1, t.v2));
+    }
+
+    default IntStream flatMapToInt(final BiFunction<? super T1, ? super T2, ? extends IntStream> mapper) {
+        return flatMapToInt(t -> mapper.apply(t.v1, t.v2));
+    }
+
+    default LongStream flatMapToLong(final BiFunction<? super T1, ? super T2, ? extends LongStream> mapper) {
+        return flatMapToLong(t -> mapper.apply(t.v1, t.v2));
+    }
+
+    default DoubleStream flatMapToDouble(final BiFunction<? super T1, ? super T2, ? extends DoubleStream> mapper) {
+        return flatMapToDouble(t -> mapper.apply(t.v1, t.v2));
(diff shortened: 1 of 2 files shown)


lukaseder:
Huh, you really wanna do this eh? :)

I like the initiative, thank you very much for this contribution. I'll think about this for a while and let you know. There are certainly some nice use-cases that can becovered by "flattening" tuples into the `Seq` type

korgov:
Yes "flattening" tuples into the Seq that I would like to get. The main reason for this is to work with Seq of Tuples without Tuples =) (without not understandable field-names like ``v1`` and ``v2``).
For example: 
after ``Seq.zip()``  we have  ``Seq<Tuple2>`` and then our code can look like ``seq.map(t -> doSomething(t.v1, t.v2)``
If we use ``Seq2.zip()`` we will be able to write more readable code like ``seq.map((id, name) -> doSomething(id, name))``

Thank you! =)

tlinkowski:
What I'd need most would be `filter1`, `map2`, etc. - the `map` method is the most cumbersome to work with since you have to manually create a new tuple each time (which is boilerplate code).

lukaseder:
@tlinkowski Indeed, I see the pain point. It is unfortunate that `Seq<T>` cannot use "higher kinded generic constraints" to introspect what `T` really is. I'm still a bit reluctant to add new `Seq[N]` types. It seems like we'll be getting a lot of type overhead...

korgov:
> I'm still a bit reluctant to add new Seq[N] types. It seems like we'll be getting a lot of type overhead...

Maybe only "Seq2" or "EntrySeq" ? Now I'm not imagine where Seq3 can be useful.

tlinkowski:
Personally, I see very little use for `Seq3` and practically no use for further `Seq[N]`. Even with tuples, I can perhaps imagine using `Tuple5` or even `Tuple6` but `Tuple16`? :) It seems to me that tuples larger than, say, 6 are only cluttering the code (but maybe I miss some use case).

So if it were up to me, I'd just add `Seq2` and focus on enhancing it with a plenty of methods, e.g. `allMatch/anyMatch/noneMatch(BiPredicate)`, `append(Map)`, `appendTuple(T1,T2)`, `contains(T1,T2)`, `distinct(BiFunction)`, `distinct1/2(Function)`, `findFirst(BiPredicate)`, `flatMap1/2(Function)`, `flip()` (would change `Seq2<T1,T2>` to `Seq2<T2,T1>`), `get1/2()`, `sorted1/2()`, etc.

lukaseder:
I see you're not working with SQL often enough :) They say MS Access's limit of 256 columns is annoying. Now imagine SQL with only 16 allowed columns?

Thanks for your comments. Will think about them

tlinkowski:
Thanks for your responses too :)

Indeed, if you mean "non-functional" tuples like mentioned SQL record holders (or perhaps, in some "Reflection" use case, holders of some class fields), I must admit that even 16 is too few - so I get back my words about tuples in general.

However, if we think about all the "functional" methods (all the `concat`, `limit`, `skip`, `map` methods in `Tuple[N]` + all the `windows` and `zip` methods in `Seq`), it still seems to me there's practically no use for them for large N, or I still miss something :)

lukaseder:
> if you mean "non-functional" tuples like mentioned SQL record holders

Well... :) In my opinion, SQL is the ultimate functional programming language (even declarative). Everything is structurally typed, not nominally typed. Which is the main reason why tuples become large. But anyway. No one uses tuples as such in SQL (attribute access by index). Everyone uses them as maps (attribute access by name).

Anyway, back to jOOλ, where 2 or 3 is often enough: I don't know :). As far as `window` methods are concerned, 16 is certainly a reasonable value, if you really want to use jOOλ for sophisticated analytics. It's not uncommon to have a dozen distinct window specifications in SQL reporting as well. But of course, that's one of the lesser used features in jOOλ. I might agree with `map` et al.

tlinkowski:
> In my opinion, SQL is the ultimate functional programming language
Of course it is :) For example, .NET's LINQ (to me, a kind of predecessor of Streams in Java) is based on SQL. Still, as you said, columns in databases are referred to by name and not by index so this specific "index-based functional approach" (`mapN`, etc.) is not used here.

As for `window`, I must admit I'm barely acquainted with this concept so I trust your words :) I seem to understand the simple usages, though, like in `Seq.sliding` method which is indeed very useful at times.

Getting back to the topic, if you seriously think about generating all `Seq[N]` interfaces, you might consider generating them up to certain N only (3 or 4?). I know that `Seq.zip` methods would be inconsistent then (some would return `Seq[N]`, and some `Seq`). However, `Seq.unzip` methods already are kind of inconsistent (they operate on `Tuple2` only). Besides, `Seq[N]` for large N would probably be used so rarely (if at all) that consistency looks like the one and only reason to generate all `Seq[N]`...

lukaseder:
> As for window, I must admit I'm barely acquainted with this concept so I trust your words :) I seem to understand the simple usages, though, like in Seq.sliding method which is indeed very useful at times.

It's not the same. "Window" here means the same as it does in SQL. The `sliding()` method also refers to the term window, but SQL window functions are much more powerful. More info here:
https://blog.jooq.org/2016/01/06/2016-will-be-the-year-remembered-as-when-java-finally-had-window-functions

In any case. It'll have to be well thought through. It may well be that `Seq[N]` for `N>1` will have much less features than `Seq` itself.

lukaseder:
OK, let's look through a few existing `Seq` methods and how they would translate in a `Seq2` type:

```java
interface Seq2<T1, T2> /* we cannot extend any type here, unfortunately */ {
    // "obvious" ones (convenience overloads are implied):
    Seq2<T1, T2> append(Seq2<? extends T1, ? extends T2> seq);
    <U1, U2> Seq2<U1, U2> cast(Class<? extends U1> c1, Class<? extends U2> c2);
    Seq2<T1, T2> concat(Seq2<? extends T1, ? extends T2> seq);
    contains(T1 t1, T2 t2);
    containsAll(Seq2<? extends T1, ? extends T2> seq);
    containsAny(Seq2<? extends T1, ? extends T2> seq);
    <U> Seq2<T1, T2> distinct(BiFunction<? super T1, ? super T2, ? extends U> keyExtractor);
    Seq2<T1, T2> filter(BiPredicate<? super T1, ? super T2> predicate);
    void forEach(BiConsumer<? super T1, ? super T2> consumer);
    Seq2<T1, T2> intersperse(T1 v1, T2 v2);
    Seq3<T1, T2, Long> zipWithIndex();
    <U> Seq3<T1, T2, U> zip(Seq<? extends U> other);

    // Reverting to Tuple2
    Tuple2<T1, T2> findFirst(BiPredicate<? super T1, ? super T2> predicate);
    Tuple2<T1, T2> findSingle();
    <K> Map<K, List<Tuple2<T1, T2>>> groupBy(BiFunction<? super T1, ? super T2, ? extends K> grouping);
    Seq2<T1, T2> iterate(T1 seed1, T2 seed2, BiFunction<? super T1, ? super T2, Tuple2<T1, T2>> iterator);
   

    // Open for discussion:
    <U> Seq3<T1, T2, U> crossJoin(Seq<? extends U> seq);
    <R> Seq<R> flatMap(BiFunction<? super T1, ? super T2, ? extends Seq<? extends R>> flatmapper);
    <U> U foldLeft(U seed, Function3<? super U, ? super T1, ? super T2, ? extends U> folder);
}

There are some things to notice here:

1. Sometimes, we will still have to revert to using `Tuple2`, because Java doesn't support multi value returns. That feels clunky, especially when the two styles are mixed as in the case of `iterate()`
2. Sometimes, there's no really good solution because we'll have to make up our minds about the degree of the resulting `Seq`, as is the case for `flatMap()`. OK, we get a `BiFunction that receives `T1` and `T2` arguments, but what do we produce? A `Seq`? Or a `Seq2`?
3. The fact that this new `Seq2` type cannot extend at least `Iterable` or `Stream` will decrease interoperability quite a bit, which is a pity.

At this stage, my reluctance to implement this feature grows :)

**Comment by korgov** [comment]
Why Seq2 cannot extend `Seq<Tuple2>` ?

1.  If it extends `Seq<Tuple2>` - It fix `Reverting to Tuple2` methods (we can use they implementation from base Seq<Tuple2>) and two styles will not be mixed.
3. If it extends `Seq<Tuple2>` - it also extends Iterable and Stream

2. For `map` / `flatMap` and other one-result methods - I think we need to produce one-type Seq (`Seq<R>`). But we can create overloaded methods like `Seq2<R,T2> map1(Function<T1, R>)` and `Seq2<R, T2> map1(BiFunction<T1, T2, R>)` - IMHO, this will be enough for those who want Seq2 after map. (Since I created this pull-request I'm using this code in my production-project and it is comfortable =) )

We can also add something like `mapToSeq2(BiFunction<T1, T2, Tuple<R1, R2>>)` but it clashes with `1st` point =)

Thanks for your response!

**Comment by lukaseder** [comment]
> Why Seq2 cannot extend Seq<Tuple2> ?

I haven't given this thorough thought, but I felt like there would be quite a bit of risk of limiting the design space of future methods, given the implications this would have on covariant method overrides in `Seq2`.

But I'll do the exercise of prototyping. Will provide more feedback.

**Comment by lukaseder** [comment]
Some examples:

```java
// These could profit from overloading, e.g Consumer<? super Tuple2<T1, T2>> 
// with Consumer2<? super T1, ? super T2>
void forEach(Consumer<? super Tuple2<T1, T2>>)
Seq<Tuple2<T1, T2>> peek(Consumer<? super Tuple2<T1, T2>>)

// These could be overridden with new types like Spliterator2, Comparator2, etc.
Spliterator<Tuple2<T1, T2>> spliterator();
Seq<Tuple2<T1, T2>> sorted(Comparator<? super Tuple2<T1, T2>>)

// These cannot be overridden covariantly, lacking declaration site variance in Java
Tuple2<Optional<Tuple2<T1, T2>>, Seq<Tuple2<T1, T2>>> splitAtHead();
Seq<Seq<Tuple2<T1, T2>>> sliding(long size);
```

Specifically the last section is really causing some headaches. If we had Tuple2<out T1, out T2> in Java, there wouldn't be any problem, but `Tuple2`, `Seq`, and `Optional` are invariant, so we cannot return, e.g.:

```java
Tuple2<Optional<Tuple2<T1, T2>>, Seq2<T1, T2>> splitAtHead();
Seq<Seq2<T1, T2>> sliding(long size);
```

... as those return types aren't subtypes of the original types in Java.

tlinkowski:
Here's what I think:
1. To me, `Seq2` is essentially a kind of `Seq`. It's like `Tuple2` is a kind of `Tuple`. Since we define inheritance as the "is-a" relationship, `Seq2` should extend `Seq`. True, things would be different if we had `Seq1` - but we don't :)
2. If `Seq2` extends `Seq`, there's no problem with overriding methods returning `Seq` to return `Seq2` (e.g. `zip`) because it'll be covariant (if `Seq2<T1,T2> extends Seq<Tuple2<T1,T2>>`).
3. As for the remaining methods: some can be (as you pointed out) overloaded (`forEach`, `peek`), some overridden (I don't think you could override `sorted` with `Comparator2`, though - it would rather be an overload), and some (the invariant ones) could simply be new methods (e.g. `splitAtHeadSeq2`, `slidingSeq2`) or they wouldn't have their `Seq2` versions at all (these methods aren't the most important anyway).
4. To sum up, it seems to me that the solution proposed by @korgov in his Pull Request is really, really good and should rather be extended and/or slightly modified than totally redesigned.

But again, I could have missed something important in my explications - if I did, please, let me know :)

lukaseder:
> 1. To me, Seq2 is essentially a kind of Seq. It's like Tuple2 is a kind of Tuple. Since we define inheritance as the "is-a" relationship, Seq2 should extend Seq. True, things would be different if we had Seq1 - but we don't :)

On the other hand, the current `Seq` *is* the `Seq1` type. In that sense, `Seq2` perhaps should extend `Stream`, rather than `Seq`. But I'm still not convinced that this "is-a" relationship is correct. By hacking this into the type system, I feel we're going to be paying a very high price in the future.

> 2. If Seq2 extends Seq, there's no problem with overriding methods returning Seq to return Seq2 (e.g. zip) because it'll be covariant (if Seq2<T1,T2> extends Seq<Tuple2<T1,T2>>).

In most cases, you're right. In some cases you're not. E.g. `zipWithIndex()` is particularly ugly.

> 3. As for the remaining methods: some can be (as you pointed out) overloaded (forEach, peek), some overridden (I don't think you could override sorted with Comparator2, though - it would rather be an overload),

Yes, I meant overloaded, not overridden. I corrected that comment.

> and some (the invariant ones) could simply be new methods (e.g. splitAtHeadSeq2, slidingSeq2) or they wouldn't have their Seq2 versions at all (these methods aren't the most important anyway).

That by itself shows that we're going the wrong way :)

> 4. To sum up, it seems to me that the solution proposed by @korgov in his Pull Request is really, really good and should rather be extended and/or slightly modified than totally redesigned.

I don't agree :) We're not quite there yet.

Let me ask you a question. Why do you think the "is-a" relationship is a good idea (or even beneficial) in the first place?

lukaseder:
(don't get me wrong. I start liking the idea of having a `Seq2` type. But I'm very much opposed to this "quick win" of letting it extend `Seq`. That just appears very short-sighted)

tlinkowski:
> On the other hand, the current Seq is the Seq1 type. In that sense, Seq2 perhaps should extend Stream, rather than Seq.

Hmm, weird... For the past hour trying to respond to your post, I alternate between feeling that `Seq` indeed is `Seq1`, and that it is not at all ;) As I cannot seem to grasp it all, I'll just write what's going on in my head.

On the one hand, `Seq` operates on all the single `T` values + all the `Function`s, `Predicate`s and `Consumer`s, which correspond to all the `T1` and `T2` value pairs + all the `BiFunction`s, `BiPredicate`s and `BiConsumer`s on which `Seq2` would operate. In this sense, `Seq` indeed is `Seq1`, and `Seq2` should _not_ extend `Seq`. Incidentally, this `Tuple2`-free operation is what seems to be the most needed feature of `Seq2`. But of course, `Tuple2`s cannot be eradicated from return types so they will be present in `Seq2` in any case.

On the other hand, it seems to me that the true `Seq1` would be defined as `Seq1<T1> extends Seq<Tuple1<T1>>` with:
`Seq<T extends Tuple> extends Iterable<T>, Stream<T> {`
`Optional<T> findSingle();`
`}`
Of course, such `Seq1` would be useless because iteration + all `Stream` operations + methods that could not be overloaded (like `findSingle`) would be `Tuple1`-based. However, it would then be consistent with what I seem to understand as the "`Seq1`-ness" ;) For example, why `Seq.window(WindowSpecification)` returns `Seq<Tuple1<Window>>`? To me, this is an example of "`Seq1`-ness" because (obviously) this method could just return `Seq<Window>` if not for the tempting consistency with other `WindowSpecification`-based methods.

To sum up, I just can't figure out right now where the boundary between generic `Seq` and single-value-specific `Seq1` lies :)

>> If Seq2 extends Seq, there's no problem with overriding methods returning Seq to return Seq2 (e.g. zip) because it'll be covariant (if Seq2 extends Seq).

> In most cases, you're right. In some cases you're not. E.g. zipWithIndex() is particularly ugly.

That's true - this one is invariant, and inheriting it from `Seq` would be really ugly indeed!

> But I'm still not convinced that this "is-a" relationship is correct. By hacking this into the type system, I feel we're going to be paying a very high price in the future.

I must admit I'm starting to share your concerns... It looks so optimistic at first sight until one gets into details (like just mentioned `zipWithIndex` or previously mentioned `splitAtHead`).

> Let me ask you a question. Why do you think the "is-a" relationship is a good idea (or even beneficial) in the first place?

I could tell you why I thought it beneficial - because of the superficial ease of implementation and interoperability (e.g. possibility of substituting `Seq2` for `Seq`). But whether it is a good idea? I'm not so confident anymore...

Well, I do feel it'd be a shame to have such a limited `Seq2` if the only step needed for it to have plenty of useful methods is to extend `Seq`. But I start to understand that succumbing to this temptation may cause regrets in the future.

So I'm now more drawn to `Seq` and `Seq2` sharing some common super interface (e.g. `BaseSeq<T extends Tuple>`) that would define a couple of "safe" common methods (like `BaseStream` for `Stream`), and for `Seq2` to have all the needed methods implemented anew (even if `Seq2` implementation would simply be a wrapper over `Seq`). But I believe that `Seq2` should then have a fairly low-cost `asSeq` (or `toSeq`) method. The opposite conversion could unfortunately be realized only by some static method on the `Seq2` interface (e.g. `Seq2.seq2`).

Perhaps, we could even venture as far as `BaseSeqN<T extends Tuple> extends BaseSeq<T>, Iterable<T>, Stream<T>` as a common super interface for `Seq2` and `Seq3`? Just an idea...

lukaseder:
Thanks for sharing your thoughts. I'm glad you're a bit more concerned about this "is-a" relationship now :) I'll reply more in depth later this week, just one more thing: `Seq2` can *look like* `Seq` without *being* `Seq`. Interoperability can be achieved with `seq()` or `stream()` methods. Once a hypothetical `Seq2` is a completely independent type (with all the functionality of `Seq`, but specialised), then the design space will become much bigger.

Java 8's `BaseStream` was a big mistake. The type is useless outside of the type hierarchy, and doesn't even add much value inside of the type hierarchy. I've blogged about this quite some time ago:
https://blog.jooq.org/2013/06/28/the-dangers-of-correlating-subtype-polymorphism-with-generic-polymorphism. Intriguing. Looks like another blog post should be written.

tlinkowski:
> Java 8's BaseStream was a big mistake. The type is useless outside of the type hierarchy, and doesn't even add much value inside of the type hierarchy. I've blogged about this quite some time ago: [...]

Well, I can certainly agree about the uselessness of this second type parameter in `BaseStream`. But if it didn't have it, and the `parallel` method simply returned `BaseStream<T>`, and then `Stream<T>` overrode it to return `Stream<T>`, this common super interface would look fine by me, even though it's not particularly useful because you could always box a primitive stream and arrive at the same place.

As for your blog post, I think it was really insightful - thanks :)

lukaseder:

However, this feature request doesn't scale very well to the currently supported 16 tuple types. I prefer not to add it.

Issue #187, opened 2016-01-16
Add support for Seq.zipAll()


lukaseder:
The current `Seq.zip()` method "inner zips" two streams, such that the size of the resulting stream is the minimum of the size of each zipped stream.

Sometimes, it is useful, however, to fully zip all streams, providing `Optional` values (or nulls) once the shorter stream is depleted. Example:

```java
// ((1, "A"), (2, "B"))
Seq.of(1, 2, 3).zip(Seq.of("A", "B"));

// ((1, "A"), (2, "B"), (3, null))
Seq.of(1, 2, 3).leftOuterZip(Seq.of("A", "B"));

// ((1, "A"), (2, "B"), (null, "C"))
Seq.of(1, 2).rightOuterZip(Seq.of("A", "B", "C"));
```

danieldietrich:
It is a special case of what Scala calls `zipAll`:

```java
interface Seq<T> {
    // thisElem and thatElem used to fill up the result if this or that runs out of elements
    <U> Seq<Tuple2<T, U>> zipAll(Seq<U> that, T thisElem, U thatElem);
}

with `thisElem` and `thatElem` set to `null`.

**Comment by lukaseder** [comment]
Yep, true. Thanks for the hint.

**Comment by danieldietrich** [comment]
I thought again about outer zips. In contrast to DB joins, zip is very limited. It composes records sequentially, one-by-one. The only match-condition the records meet is their position.

Outer joins handle the case where for a given record of one relation no record of the other relation matches the condition. zipAll handles both left-and right outer join simulaneously by providing default values. Therefore zipAll is _not_ a special case of outer zips.

**Comment by lukaseder** [comment]
Well... In DB Speak, the following can be said:

```sql
-- Convenience form
A [ { LEFT | RIGHT | FULL } [ OUTER ] ] ZIP B

-- Full form
(SELECT A.*, ROW_NUMBER() OVER() AS RN FROM A) AS A
[ { LEFT | RIGHT | FULL } [ OUTER ] ] JOIN
(SELECT B.*, ROW_NUMBER() OVER() AS RN FROM B) AS B
ON A.RN = B.RN
```

`zipAll()` is an alias for `fullOuterZip()`

danieldietrich:
Thank you. If we had functions

```java
Seq<T> continue(T next);
Seq<T> continue(Supplier<? extends T> nextSupplier);
Seq<T> continue(Function<? super T, ? extends T> previousToNext);
```

then

```java
[...]

PR #200, opened 2016-02-28
[#187] Add Seq.zipAll()

patrox:


lukaseder:
Thank you very much for this contribution. It's already a great start! I'll go and fix the code generator in order to generate also all the other 14 degrees :)

patrox:
@lukaseder you're very welcome !

> I'll go and fix the code generator in order to generate also all the other 14 degrees

That's a relief, as I thought that I'll need to add the missing ones as well ;)

lukaseder:
Yeah, those are a pain to write manually ;)

Issue #205, opened 2016-03-09
Add <T> Seq<T> seq(Enumeration<T> enumeration)


sargue:
```java
interface Seq {
    /**
     * Wrap an <code>Enumeration</code> into a <code>Seq</code>.
     */
    static <T> Seq<T> seq(Enumeration<T> enumeration) { ... }
}
```

What do you think about this? It should work like the Iterator wrapper (probably with some more boilerplate code as I think there is no standard Spliterator for Enumerations).

sargue:
Possible implementation?

```java
static <T> Seq<T> seq(Enumeration<T> enumeration) {
    Spliterator<T> spliterator = Spliterators.spliteratorUnknownSize(new Iterator<T>() {
        @Override
        public boolean hasNext() {
            return enumeration.hasMoreElements();
        }

        @Override
        public T next() {
            return enumeration.nextElement();
        }
    }, ORDERED);
    return Seq.seq(spliterator);
}
```

lukaseder:
Nice thinking. This would certainly help working with legacy APIs like `Vector` or `Properties`. Did you have any other specific use-case in mind?

Would you be interested to provide this as a pull request?

sargue:
Yeah, mostly legacy APIs. I can't think of another use. I've found myself using it with the Servlet API `HttpServletRequest.getParameterNames()` for example.

I will prepare and send a PR.

lukaseder:
Ah, true. Servlets also use this type, still. Thanks, looking forward!

lukaseder:
Oh, btw, I think you can skip the `Spliterator` and pass the `Iterator` directly to `seq()`

PR #206, opened 2016-03-10
Added Seq.seq(Enumeration) to better support legacy APIs

sargue:
Resolves jOOQ/jOOL#205


lukaseder:
Thank you very much for this!

Issue #207, opened 2016-03-11
Add method to create maps from streams with the stream objects as the map values


sargue:
I've been using this construction a lot lately:

```java
Seq<SomeObject> seq = ...;
Map<Integer, SomeObject> objectMap = seq.toMap(SomeObject::getId, Function.identity());
```

And I've think this sugar-addition would be nice for jOOL.

```java
interface Collectable<T> {
    /**
     * Collect the collectable into a {@link Map} with the given keys
     * and the self element as value.
     */
    <K> Map<K, T> toIdentityMap(Function<? super T, ? extends K> keyMapper);
}
```

I can provide a PR. Alternative names to `toIdentityMap` are very welcome. ;-)

lukaseder:
Interesting. Yes, there's room for this feature. Also, the inverse might be useful. For instance, when you have a `Seq<String>`, and you want to create a dictionary from it, like so:

```java
    <V> Map<T, V> toXXXMap(Function<? super T, ? extends V> valueMapper);
```

Now, for the hardest problem: Naming ;)

I don't think that the term "identity" should be used. This is really about keys / values. For consistency reasons, the `to` prefix will be mandatory. Same for the `Map` suffix. `toMapKeys()` and `toMapValues()` aren't perfect, because it's not clear which one is which. E.g. does `Keys` mean that keys are mapped, or that keys are kept? Hmm...

amaembo:
FYI I named such thing simply as [`toMap`](http://amaembo.github.io/streamex/javadoc/one/util/streamex/StreamEx.html#toMap-java.util.function.Function-) providing an overload when value mapper only is specified. With StreamEx, in OP case you still should specify explicitly both mappers. Actually adding `x->x` is not that long, so it's not the big problem.

lukaseder:
Hmm, I'd prefer the keymapper version if an overload is provided. Simply because with convenience overloads, usually, the last argument(s) are the optional ones:

```java
toMap(keyMapper) { return toMap(keyMapper, v -> v); }
toMap(keyMapper, valueMapper)
```

sargue:
@amaembo, I also thought about the `x->x`solution. That would be:

```java
Seq<SomeObject> seq = ...;
Map<Integer, SomeObject> objectMap = seq.toMap(SomeObject::getId, x->x);
```

Or the inverse as commented by @lukaseder 

```java
Seq<String> names = ...;
Map<String,V> dictionary = names.toMap(x->x, Converter::fromString);
```

[...]

PR #208, opened 2016-03-12
Added overloaded Collectable.toMap(keyMapper) to address the common…

sargue:
… case of creating dictionaries.

Closes jOOQ/jOOL#207


lukaseder:
Thank you very much for this!

Issue #57, opened 2014-10-03
Enhancement: replace multitude of Unchecked methods with single overloaded method


almson:
I did this in my own utils library (`LambdaUtils.unchecked`) and it works a treat.

PR #229, opened 2016-04-23
Add aliases for all functions in »Unchecked«.

ErsafanRend:
I'd like to use the functions in »Unchecked« via static import. 
Unfortunately semantics are a little bit lost when I use
    `Stream.of(blah).map(function(OpThat::throws))`
I think it is much clearer to have
    `Stream.of(blah).map(unchecked(OpThat::throws))`
Maybe this commit also solves #57.


lukaseder:
Thank you very much for your suggestion. I agree that it would be much clearer, but unfortunately, it is a very bad idea to overload methods that take lambda arguments. The rationale is explained here:
https://blog.jooq.org/2015/01/29/you-will-regret-applying-overloading-with-lambdas

lukaseder:
... I'm sorry, I was a bit quick closing that withouth having read your suggested code. With various `unchecked` methods in distinct classes, this issue wouldn't arise. What is the reason why you put those methods in new classes rather than in the existing functional interfaces?

ErsafanRend:
> What is the reason why you put those methods in new classes rather than in the existing functional interfaces?
Actually I didn't think that far. Putting the unchecked methods as default implementations in the interfaces would be much more elegant. If you prefer I will migrate them.

lukaseder:
Not as default methods, but as static methods... Yes, that would certainly be adding some value to the `Unchecked` API, greatly appreciated!

ErsafanRend:
How do you think about moving the implementations from Unchecked to the functional interfaces?
So to speak a swap of alias and implementation?

Pro:
:+1: Implementations are separated in their interfaces.
:+1: The only dependency of the functional interfaces is towards the consumers in »Unchecked«.

Con:
:-1: Similar but somehow duplicated doc blocks in the functional interfaces as each »unchecked« method needs its own example.

lukaseder:
Thank you very much for the update. I've merged this as it stands right now. Will comment on your suggestion [here](https://github.com/jOOQ/jOOL/pull/229#issuecomment-213796334) in a bit

lukaseder:
Hmm, no, I wouldn't want to swap the implementations and the delegations. I like the fact that all the implementations are in one place. It seems easier to change things in the future, this way.

ErsafanRend:
That's fine for me. 

And thank you for merging the PR. I really like the leightweight interface that jOOλ provides.

Issue #213, opened 2016-03-13
Add Seq.seq(Optional<? extends T>...)


lukaseder:
In addition to the existing `seq(Optional<? extends T>)`. This goes along well with the existing `concat(Optional<? extends T>)`

lukaseder:
Discussion in the PR: https://github.com/jOOQ/jOOL/pull/231

PR #231, opened 2016-05-21
Seq.seq(Optional<? extends T>...) implementation

earlzero:
Implementation for #213.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -4592,6 +4592,14 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
         return optional.map(Seq::of).orElseGet(Seq::empty);
     }
 
+
+    /**
+     * Wrap multiple <code>Optional</code>'s into a <code>Seq</code>.
+     */
+    static <T> Seq<T> seq(Optional<? extends T>... optionals) {
+        return seq(Arrays.stream(optionals).flatMap(o -> o.isPresent() ? Stream.of(o.get()) : Stream.empty()));
+    }
+
     /**
      * Wrap an <code>InputStream</code> into a <code>Seq</code>.
      * <p>
--- a/src/test/java/org/jooq/lambda/SeqTest.java
+++ b/src/test/java/org/jooq/lambda/SeqTest.java
@@ -1392,6 +1392,12 @@ public class SeqTest {
     }
 
     @Test
+    public void testOptionals() {
+        assertEquals(asList(1), Seq.seq(Optional.of(1), Optional.empty()).toList());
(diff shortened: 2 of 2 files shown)


amaembo:
At least I'd suggest to replace `Stream.empty()` with `null` which is allowed for `flatMap`, means the same and works notably faster. But it would be even faster to use `.filter(Optional::isPresent).map(Optional::get)` as no intermediate streams are created here.

tkroman:
@earlzero, would you consider adding an instance method for this as well (if there's none)?
Something that works like this: ` flattenOptionals: Seq<Optional<? extends T>> -> Seq<? extends T> `

lukaseder:
> At least I'd suggest to replace Stream.empty() with null

o_O, how could that implementation detail possibly leak into the public API?? Oh my... `Stream.concat()` should allow null arguments too, then?

lukaseder:
@earlzero : Thank you very much for your implementation suggestion. The reason why this hasn't been implemented yet is because of nasty warnings at the call-site about heap pollution. I'm not 100% convinced yet, that this should really be added.

earlzero:
@lukaseder don't mention. Anyway it was pretty  straightforward one :).

tkroman:
@lukaseder But this would be sooo useful! You should try to convince yourself that this is needed :D

lukaseder:
@earlzero @tkroman We're not in a hurry to make a decision. I just wanted to state why this hasn't been implemented yet. :)

earlzero:
@lukaseder I checked why heap pollution may occur (https://docs.oracle.com/javase/tutorial/java/generics/nonReifiableVarargsType.html#vulnerabilities). I don't think it's applicable to this case (we don't rely on features of the type T, we don't try to change optionals array). I can use `SafeVarargs` to prevent this message from appearing, if I convinced you enough :).

@amaembo filter/map solution will definitely work faster, but I don't think this particular place is performance critical.

lukaseder:
@earlzero I know that heap pollution cannot occur in this case, but there's still a nasty warning at the client side. Unfortunately, the `@SafeVarargs` annotation cannot be placed on interface methods...

earlzero:
@lukaseder SafeVarargs can't be put on non-final or non-static methods, but seq here is static, so we can use it.

lukaseder:
My bad, I overlooked that detail. I'll merge your suggestion, apply the fix suggested by @amaembo, and add `@SafeVarargs`

lukaseder:
Thanks again!

lukaseder:
In fact, why not just get rid of `flatMap` entirely and do this instead:

```java
    static <T> Seq<T> seq(Optional<? extends T>... optionals) {
        return of(optionals).filter(Optional::isPresent).map(Optional::get);
    }
```

amaembo:
That's exactly what I suggested!

lukaseder:
Oh, I TL;DR your reply, I suspect :)

PR #234, opened 2016-06-13
Wrap IOException with UncheckedIOException in Unchecked

Xaerxess:
JDK8 comes with [this gem](https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html), so it seems natural that `Unckecked` should wrap most hated Java checked exception with new unckecked counterpart.

### DIFF
--- a/src/main/java/org/jooq/lambda/Unchecked.java
+++ b/src/main/java/org/jooq/lambda/Unchecked.java
@@ -20,6 +20,8 @@ import org.jooq.lambda.fi.util.function.*;
 import org.jooq.lambda.fi.lang.CheckedRunnable;
 import org.jooq.lambda.fi.util.CheckedComparator;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Comparator;
 import java.util.function.*;
 
@@ -46,6 +48,9 @@ public final class Unchecked {
 
         if (t instanceof RuntimeException)
             throw (RuntimeException) t;
+
+        if (t instanceof IOException)
+            throw new UncheckedIOException((IOException) t);
         
         // [#230] Clients will not expect needing to handle this.
         if (t instanceof InterruptedException)


ansell:
+1 from me to following JDK conventions where they exist. Great find!

src/main/java/org/jooq/lambda/Unchecked.java
```diff
@@ -46,6 +46,9 @@
 
         if (t instanceof RuntimeException)
             throw (RuntimeException) t;
+
+        if (t instanceof IOException)
+            throw new UncheckedIOException((IOException) t);
```
ansell: UncheckedIOException and IOException need import statements

Xaerxess: Of course, my bad. IntelliJ changed `import java.util.function.*` to bunch of imports, so I ignored them when doing `git add -p` ;)

lukaseder:
Good idea! I suspect the Javadoc might need some adaptations, too. Would you be interested in doing that? Otherwise I will. There had been another recent change in this class that should be documented.

Xaerxess:
@lukaseder What do you specifically have in mind? `THROWABLE_TO_RUNTIME_EXCEPTION` documentation or another methods?

lukaseder:
@Xaerxess Yes, that would be a good place. There's already some mention of this `THROWABLE_TO_RUNTIME_EXCEPTION` being applied by default, so it would be a good place to put any documentation, to keep things "DRY"

Xaerxess:
@lukaseder Regarding documentation in `Unchecked` - should `THROWABLE_TO_RUNTIME_EXCEPTION` be considered as default for 1.0 or maybe sneaky throws (`RETHROW_ALL`) should be used? [Its behavior is in JLS](http://stackoverflow.com/a/31316879/708434), so it's quite elegant IMO.

lukaseder:
I don't think sneaky throw is a good default as it is very treacherous. Some checked exceptions really convey meaning in terms of being business exceptions (unlike `SQLException` or `IOException` which are usually system exceptions, i.e. good runtime exception candidates). With that in mind, it is not possible to catch a checked exception from a statement that does not declare throwing it. Try:

```java
try {
    System.out.println();
}
catch (SQLException e) {
}
```

Of course, we could simply catch-all and do instanceof checks, but I feel like this is a dirty approach...

Besides, the class is called `Unchecked`, so rethrowing unchecked exception is what's expected. Perhaps, we can create a new class that implements the desired behaviour as a default...?

Xaerxess:
> With that in mind, it is not possible to catch a checked exception from a statement that does not declare throwing it.
That sounds very reasonable, catch-all with instanceof in sneaky case seems really dirty. Let's just stick to current handler then. 

> Perhaps, we can create a new class that implements the desired behaviour as a default...?
So now just find a name for Unchecked equivalent with sneaky throw exception handler ;)

But back to the topic: I'll try to upgrade `THROWABLE_TO_RUNTIME_EXCEPTION` documentation a bit (in a separate PR).

lukaseder:
> So now just find a name for Unchecked equivalent with sneaky throw exception handler ;)

Yeah. Almost impossible :P

> But back to the topic: I'll try to upgrade THROWABLE_TO_RUNTIME_EXCEPTION documentation a bit (in a separate PR).

Great, thanks very much!

lukaseder:
Hmm, in fact, this issue is already fixed... (original improvement: `UncheckedIOException`)

PR #240, opened 2016-08-04
Blocking#runnable() was not executing the runnable given by argument.

gortiz:
The previous implementation of Blocking#runnable(Runnable) created a
blocking supplier, but it did not call Supplier#get(), so the runnable
given by argument was never executed.


lukaseder:
Oh my, thanks a lot! Nice catch

Issue #242, opened 2016-08-12
Add Seq.indexOf method(s)


tlinkowski:
Since we have `Seq.contains` and `Seq.get(index)`, it'd be nice to also have `long indexOf(T item)` or `OptionalLong indexOf(T item)`.

Incidentally, we could also have `long/OptionalLong indexOf(Predicate<? super T> predicate)` which would be equivalent to Guava's `Iterators.indexOf`.

lukaseder:
Excellent idea. I agree with both proposed APIs:

```java
OptionalLong indexOf(T item)
OptionalLong indexOf(Predicate<? super T> predicate)
```

The latter is also not unexpected, given the existing `Seq.findFirst(Predicate)`. I think we should prefer `OptionalLong` over the historic `-1` encoding in a primitive `long` as it fits better into the `Stream` / `Seq` API.

Would you be interested in providing a pull request for this?

amaembo:
I have such [methods](http://amaembo.github.io/streamex/javadoc/one/util/streamex/AbstractStreamEx.html#indexOf-T-) returning `OptionalLong`. Returning -1 seems not consistent with other Stream API methods. Though probably a matter of taste.

tlinkowski:
I am also in favor of `OptionalLong`.

OK, I'll create a pull request for this over the weekend. I propose an implementation utilizing `iterator()` and mirroring Guava's `Iterators.indexOf`.

lukaseder:
Excellent, looking forward!

lukaseder:
Implementation in #243

PR #243, opened 2016-08-13
[#242] Add Seq.indexOf method(s)

tlinkowski:
See #242.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -44,6 +44,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


lukaseder:
Looks good, thank you very much!

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -2981,6 +2996,18 @@ public R next() {
     }
 
     /**
+     * Get the index of the first element matching the given predicate.
+     */
+    static <T> OptionalLong indexOf(Iterator<T> iterator, Predicate<? super T> predicate) {
```
lukaseder: I would prefer not to make this method publicly available, as I'm not 100% sure about the API design yet. So, I'll move it to `SeqImpl`...

PR #244, opened 2016-08-13
Share IntelliJ's codeStyleSettings.xml and inspectionProfiles/

tlinkowski:
IntelliJ gave me a hard time trying to reformat the `Seq` interface using the default formatting rules (for example, it tried to optimize the imports and hanged for nearly a minute).

Would you mind sharing your `codeStyleSettings.xml` and, optionally, `inspectionProfiles` so that it's easier to provide potential future pull requests?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
-.idea
+/.idea/*
+!/.idea/inspectionProfiles/
+!/.idea/codeStyleSettings.xml
 target
 *.iml


tlinkowski:
It also seems that the source code is not uniformly formatted (e.g. there are trailing spaces at some lines, and sometimes the indentation is inconsistent).

It'd be nice if the entire code were reformatted according to the rules in `codeStyleSettings.xml` that you'll (hopefully) share.

lukaseder:
I'm using NetBeans for this project, because IntelliJ is too slow with the amount of generics exposed here.

I don't agree with auto reformatting in general. Specifically with code that is as fluent as Stream / Seq usage (or jOOQ as well), auto formatting creates more damage than good. In particular, there is tons of bikeshedding potential with these rules and which ones to agree upon.

I agree that trailing spaces and inconsistent indentation (only 4 spaces should be used, no tabs) are a flaw, but then again, there's more important things to spend time on :)

tlinkowski:
OK, I thought you were using IntelliJ because I saw `.idea` in `.gitignore` and no trace of `nbproject` there (I used to code in NetBeans but now I really appreciate IntelliJ for its "implicitness").

As for IntelliJ's slowness - I am no expert but I was sure it's the sheer size of some single files (like `Seq`) rather than the size of the entire project. I work with IntelliJ on projects containing hundreds thousands of LOC but the longest Java file has around a thousand LOC, and I experience no lags whatsoever. But with the jOOL project, IntelliJ actually crashes on me quite regularly (but I guess it is partly because of its autoformatting attempts).

I must say I disagree that autoformatting causes more damage than good (I believe that quite the contrary is true). As for agreeing on which autoformatting rules to use, it doesn't matter to me much - I just appreciate that there is some setting in the IDE because then I don't have to think about this too much (IDE does most of it for me).

However, I agree that there's plenty of much more important things to do than worry about formatting :) (but again, that's why I like my IDE to handle that :P)

lukaseder:
> I must say I disagree that autoformatting causes more damage than good (I believe that quite the contrary is true)

It depends. But the discussion usually isn't very constructive. And as you've seen, we don't even know what "standard" to use to store these things. So I'd rather simply not :)

PR #250, opened 2016-08-22
Use binarySearch in rank aggs

Xaerxess:
I removed couple TODOs in `rankBy`, `denseRankBy` and `percentRankBy` aggs which now use more optimized versions. First two now use `Collections.binarySearch` with a simple while to determine first occurrence (position) in list, last one uses dedicated method in `TreeSet` since it was already used.

### DIFF
--- a/src/main/java/org/jooq/lambda/Agg.java
+++ b/src/main/java/org/jooq/lambda/Agg.java
@@ -651,7 +651,7 @@ public class Agg {
         );
     }
 
-    private static <U> int binarySearchRank(ArrayList<U> l, U value, Comparator<? super U> comparator) {
+    private static <U> int binarySearchRank(List<? extends U> l, U value, Comparator<? super U> comparator) {
         int index = Collections.binarySearch(l, value, comparator);
         if (index < 0)
             index = -index - 1;
--- a/src/main/java/org/jooq/lambda/Agg.java
+++ b/src/main/java/org/jooq/lambda/Agg.java
@@ -644,13 +644,14 @@ public class Agg {
                 if (size == 0)
                     return Optional.empty();
 
-                // TODO: Find a faster implementation using binarySearch
                 l.sort(comparator);
-                for (int i = 0; i < size; i++)
-                    if (comparator.compare(value, l.get(i)) <= 0)
-                        return Optional.of((long) i);
-
-                return Optional.of((long) size);
+                int index = Collections.binarySearch(l, value, comparator);
+                if (index < 0)
+                    index = -index - 1;
+                else
+                    while (index > 0 && l.get(index) == l.get(index - 1))
+                        --index;
+                return Optional.of((long) index);
             }
         );
     }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -743,13 +737,14 @@ else if (compare > 0)
...
-                    if (comparator.compare(value, l.get(i)) <= 0)
-                        return Optional.of((double) i / (double) size);
-
-                return Optional.of(1.0);
+                int index = Collections.binarySearch(l, value, comparator);
+                if (index < 0)
+                    index = -index - 1;
+                else
+                    while (index > 0 && l.get(index) == l.get(index - 1))
+                        --index;
```
Xaerxess: @lukaseder What's your opinion on copying 6 lines above in `rankBy` and `denseRankBy`? Should I place `firstIndexOf` somewhere? I don't see any appropriate util class for that, so please advise :)

lukaseder: Good point, how about making it a private method directly in `Agg`?

lukaseder:
Thank you very much for your contribution. I'll review it later this week.

src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -644,17 +644,23 @@ else if (compare > 0)
...
-                        return Optional.of((long) i);
-
-                return Optional.of((long) size);
+                int index = binarySearchRank(l, value, comparator);
+                return Optional.of((long) index);
             }
         );
     }
 
+    private static <U> int binarySearchRank(List<? extends U> l, U value, Comparator<? super U> comparator) {
```
Xaerxess: I created this method and chose name `binarySearchRank` because:
1) it requires list to be sorted just like `Collections.binarySearch`,
2) it does not return negative indices like original, but rather returns positive rank.

Xaerxess:
This approach could be O(n) in worst case (all values are the same) and modified binary search can be written (one that would continue searching without linear iteration), but it would require more custom code, so I felt re-using JDK's `Collections.binarySearch` would be better here.

amaembo:
Am I understand correctly that `collect(rankBy(val, cmp))` is very close to `filter(x -> cmp.compare(x, val) < 0).count()` (the difference is some special handling of empty input stream)? Am I missing something?

lukaseder:
> [ @amaembo ] Am I understand correctly that collect(rankBy(val, cmp)) is very close to filter(x -> cmp.compare(x, val) < 0).count()

The function is inspired by SQL's `RANK(val) WITHIN GROUP (ORDER BY cmp)` ordered-set aggregate function. Indeed, your alternative using `filter()` and `count()` does something similar, except that it's not a reusable `Collector`, of course.

But of course, your input shows that the existing canonical solution is a bit dumb from a performance perspective. We shouldn't collect all the values and sort the result. We should only just maintain a counter in the collector.

Xaerxess:
> We shouldn't collect all the values and sort the result. We should only just maintain a counter in the collector.
True, other collectors in `Agg` could be optimized this way, I'd open separate ticket for creating specialized collectors. Here I just found TODO which I fixed, I didn't question current approach collecting values.

lukaseder:
> True, other collectors in Agg could be optimized this way, I'd open separate ticket for creating specialized collectors. Here I just found TODO which I fixed, I didn't question current approach collecting values.

Sure, no worries! Thanks for your attempt. I'll close this issue and open a new one for the improved improvement

lukaseder:
Replacement issue: https://github.com/jOOQ/jOOL/issues/254

lukaseder:
> [ @Xaerxess ] True, other collectors in Agg could be optimized this way, I'd open separate ticket for creating specialized collectors.

OK, I improved `rankBy()` and `denseRankBy()` in #254. If you're interested, I'm sure there are also improvements possible for `percentRankBy()` and others :)

PR #251, opened 2016-08-22
Unchecked.objIntConsumer doesn't throw IllegalStateException as it should

Xaerxess:
I added missing `IllegalStateException` in `Unchecked.objIntConsumer` (I guess it wasn't auto-generated). It wasn't caught by tests - unfortunately they use ISE as expected exception, so in order to add test for this case refactoring is needed. Did you consider using parametrized tests to cover all variants of methods (not only in `Unchecked`)?

Two other minor changes:
 - removed redundant `final`,
 - used method reference - I wonder if you fancy those, because I don't see many method references in the code.

### DIFF
--- a/src/main/java/org/jooq/lambda/Unchecked.java
+++ b/src/main/java/org/jooq/lambda/Unchecked.java
@@ -67,7 +67,7 @@ public final class Unchecked {
     /**
      * "sneaky-throw" a checked exception or throwable.
      */
-    public static final void throwChecked(Throwable t) {
+    public static void throwChecked(Throwable t) {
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your contribution. I agree with the first two commits, will revert the removal of `final`, though. While it is redunant, I just like as many methods being final as possible. Who knows if we open up `Unchecked` in the future.

Parameterised tests are a good idea. Fancy another PR? :)

lukaseder:
Hmm, OK, the other methods aren't final either. I guess I've been doing that only in jOOQ... Will leave it as it is now.

PR #252, opened 2016-08-22
Add example with method references for Unchecked

Xaerxess:
Because *I* fancy method references and long stream chains ;)


lukaseder:
Yeah, nice! Thank you very much!

Issue #254, opened 2016-08-22
More optimal Agg.rankBy() and Agg.denseRankBy() implementations shouldn't keep buffers and sort them


lukaseder:
The current implementations of `Agg.rankBy()` and `Agg.denseRankBy()` are very wasteful as they:

- maintain an in-memory buffer of all the values
- sort that buffer
- run through that buffer to find the ranked index

There was a suggested pull request (https://github.com/jOOQ/jOOL/pull/250) to optimise the search for the index, but a much better approach was suggested by @amaembo (https://github.com/jOOQ/jOOL/pull/250#issuecomment-241474427): Just count the elements that are lower than the argument value

Xaerxess:
Regarding `percentRankBy` - one can implement it just like `rankBy` but with second long storing count (`() -> new long[] { -1L, 0L }` and so on), or in terms of `rankBy` using other stuff from jOOL:

    public static <T, U> Collector<T, ?, Optional<Double>> percentRankBy(U value, Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
        return collectingAndThen(
                Tuple.collectors(
                        rankBy(value, function, comparator),
                        count()),
                t -> t.map((rank, count) -> rank.map(r -> (double) r / count))
        );
    }

Which approach do you prefer?

BTW: Why are you using one-element array everywhere, why not implement mutable lazy holder class (mutable optional?) which would take care of some repeating boilerplate code? (Genuine question - I wonder if you thought about that; using l[0] looks like the hard way to me.)

lukaseder:
> Regarding percentRankBy - one can implement it just like rankBy but with second long storing count (() -> new long[] { -1L, 0L } and so on)

Excellent.

> Which approach do you prefer?

The `Tuple.collectors()` approach. Nice thinking

> BTW: Why are you using one-element array everywhere

I was inspired by the JDK, e.g.: https://github.com/dmlloyd/openjdk/blob/c9f3c7dda232dabcfe297486d8d2f341bcb1e901/jdk/src/java.base/share/classes/java/util/stream/Collectors.java#L622

Without having checked, this approach *probably* has the advantage of:

- No indirection through abstraction, because by default, the JIT inlines only 9 method levels, if I remember correctly. With arrays, we're not wasting one level on something trivial.
- Less object size overhead on pairs of `long`, and *probably* (didn't verify this) better access patterns

I'm sure when Valhalla reaches GA, all of these arrays will be replaced by actual value types.

[...]

PR #256, opened 2016-08-23
Implement Agg.percentRankBy() in terms of Agg.rankBy()

Xaerxess:
As per [comment](https://github.com/jOOQ/jOOL/issues/254#issuecomment-241574426) in PR #254 here comes `percentRankBy` implementation using `Tuple.collectors` and `rankBy`.


amaembo:
What about division by zero?

Xaerxess:
Should not occur - `Optional` would be empty (from `rankBy`) and map should
not execute.

23 sie 2016 19:35 "Tagir Valeev" <notifications@github.com> napisał(a):

amaembo:
Looks like you're right. Ok then.

lukaseder:
Thank you very much!

PR #262, opened 2016-09-22
Add Seq#zipWithIndex(BiFunction)

Xaerxess:
I wanted to use `Seq#zipWithIndex(BiFunction)`, but unfortunately it wasn't available, so here's implementation. Instead writing:

    Seq.of("a", "b", "c")
            .zipWithIndex()
            .map(t -> t.map((s, i) -> i + ":" + s))
            .forEach(System.out::println);

one can write:

    Seq.of("a", "b", "c")
            .zipWithIndex((s, i) -> i + ":" + s))
            .forEach(System.out::println);

(and no `Tuple2` object is created.)

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -1049,7 +1049,7 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
      * Seq.of(1, 2, 3).zip(Seq.of("a", "b", "c"), (i, s) -> i + ":" + s)
      * </pre></code>
      *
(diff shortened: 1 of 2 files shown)


lukaseder:
Nice idea, thank you very much for the suggestion!

Xaerxess:
Note that with `Seq2` (PR #172) it'd be cleaner even without this PR (although tuple objects still would be created).

lukaseder:
I don't think this is unclean. And this `Seq2` thing is still giving me headaches :)

Xaerxess:
Just sayin' :) And thanks for quick merge!

PR #266, opened 2016-11-18
Agg filter cleanup

Xaerxess:
As a follow-up to [this SO thread](http://stackoverflow.com/questions/34621835/java-8-streams-how-to-call-once-the-collection-stream-method-and-retrieve-an/34622098#comment68470764_34622098), I'd like to make the code simpler and yet doing the same ;)


lukaseder:
Man, I must've been tired. Thank you very much ! :)

PR #267, opened 2016-11-19
Add Agg#first() and Agg#last() collectors

Xaerxess:
I implemented `first()` and `last()` collectors, which seems like reasonable feature with `Tuple.collectors` (I had use case where I needed `first()`). [`StreamEx` has those collectors](https://github.com/amaembo/streamex/blob/streamex-0.6.3/src/main/java/one/util/streamex/MoreCollectors.java#L511-L549), but there are two differences with my implementation:

1. StreamEx's `MoreCollectors#first()` can be short circuiting - in jOOL I don't see a purpose for that.
2. Both `first()` and `last()` throw NPE when first / last element is `null` - this is to distinguish empty stream from null element, and also follows [`Stream.findFirst()`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#findFirst--) convention.

### DIFF
--- a/src/main/java/org/jooq/lambda/Agg.java
+++ b/src/main/java/org/jooq/lambda/Agg.java
@@ -57,6 +57,40 @@ public class Agg {
     }
 
     /**
+     * Get a {@link Collector} that calculates the <code>FIRST</code> function.
+     *
+     * @throws NullPointerException if first element is {@code null}
+     */
+    public static <T> Collector<T, ?, Optional<T>> first() {
+        final Object empty = new Object();
+        return Collector.of(
+                () -> new Object[] { empty },
+                (a, t) -> {
+                    if (a[0] == empty) {
+                        a[0] = t;
+                    }
+                },
+                (a1, a2) -> a1,
+                a -> a[0] == empty ? Optional.empty() : Optional.of((T) a[0])
+        );
+    }
+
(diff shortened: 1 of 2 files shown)


Xaerxess:
Ad. 2. I gave it a second thought and based on `Collectors.reducing()` vs `Stream.reduce()` behavior I think we shouldn't throw NPE in this case (as it's not direct equivalent to `Stream.findFirst()`).

@lukaseder Which approach do you think is better?

lukaseder:
Thank you very much for your suggestions. I agree that such collectors will be useful, occasionally. Note that there's already a way to access `firstValue()` and `lastValue()` through `window()` as well, although they're not grouping/collecting.

SQL (at least Oracle SQL) also has `FIRST` and `LAST` functions, so I also agree with the name (although in SQL, they work subtly differently).

> @lukaseder Which approach do you think is better?

I really prefer there not to be any NPEs. It's a shame that Java 8 introduced even more ambiguity with respect to `null` and what best practices are. I wish the Java folks would finally accept that `null` is a real thing and we'll live with it.

Thus, I definitely prefer your approach that follows `reducing` here.

Xaerxess:
Thanks for merging! Regarding SQL part - I won't question you at all ;)

Issue #225, opened 2016-03-21
Add CheckedConsumer[N] and CheckedFunction[N] and add wrapper methods to Unchecked


lukaseder:
See also: https://github.com/jOOQ/jOOL/issues/214#issuecomment-198286611

bendem:
Came here to say that. I just had to write
```java
.map(envelopeAndPath -> envelopeAndPath.map2(Function1.from(Unchecked.function(Files::newInputStream))))
```
Expected to find a method allowing this instead:
```
.map(envelopeAndPath -> envelopeAndPath.map2(Unchecked.function1(Files::newInputStream)))
```

lukaseder:
Thanks for the feedback, @bendem. This will definitely add value when implemented.

patrox:
hi @lukaseder & @bendem, could you please have a look at the following PR and let me know if this is what you were after ?

https://github.com/jOOQ/jOOL/pull/272

Please note though, that for now this has been done for arity of 1.

lukaseder:
Hmm, I'm having a look at this again. I'm not quite sure what's wrong with this, @bendem:

```java
.map(envelopeAndPath -> envelopeAndPath.map2(Unchecked.function1(Files::newInputStream)))
```

`Tuple2.map2(Function)` should accept the `Function` returned from `Unchecked.function()`. There's no need (IMO) to wrap this in a `Function1`, no?

Thanks for your suggestion, @patrox. That might be along the lines of what's being discussed here, although I wonder if we cannot just overload the existing `function()` and `consumer()` static utility methods in `Unchecked` and `Sneaky`. As the `Function[N]` and `Consumer[N]` arguments have different arities, the overload resolution should be non-ambiguous, no?

Do note that this code is generated, just as a hint before you move on to suggest manual implementations for all degrees 😉

patrox:
@lukaseder yep, I'm aware that other arity methods are generated, so that's why I created just one set for the single arity to show the general idea :)

lukaseder:
@patrox: OK, I see. So, before discussing PRs, let's discuss use-cases. How would you like to use this new feature?

bendem:
Yes, my specific problem was fixed by https://github.com/jOOQ/jOOL/commit/91e719d1256674a0eef20d179d08a260b9cf75c2. Before that you had to use jool's `Function1`.

lukaseder:
Oh, I see. Thanks for the feedback. I hadn't checked the dates...

PR #272, opened 2016-12-01
add Checked{Function, Consumer}1 and wrapper methods for Unchecked & Sneaky

patrox:


lukaseder:
As discussed in #225: #225 still needs investigation. Plus, the code is generated, so I'm currently not merging this. But thank you anyway :)

PR #278, opened 2016-12-16
Add Unchecked.callable and friends

xathien:
I couldn't find an Issue to match this description. Should I create one?

### DIFF
--- a/src/main/java/org/jooq/lambda/Sneaky.java
+++ b/src/main/java/org/jooq/lambda/Sneaky.java
@@ -19,8 +19,10 @@ package org.jooq.lambda;
 import org.jooq.lambda.fi.util.function.*;
 import org.jooq.lambda.fi.lang.CheckedRunnable;
 import org.jooq.lambda.fi.util.CheckedComparator;
+import org.jooq.lambda.fi.util.concurrent.CheckedCallable;
 
 import java.util.Comparator;
+import java.util.concurrent.Callable;
 import java.util.function.*;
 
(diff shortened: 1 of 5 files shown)


lukaseder:
Thank you very much for your suggestion. The reason why it's not in there is because `Callable` already allows for throwing checked `Exception` (granted, not `Throwable`, but that's usually not too much of a limitation).

What would be a compelling reason to still add a "CheckedCallable"? Symmetry, perhaps?

xathien:
For my projects, it is indeed so I can deal with `Throwable` in a significantly prettier way. A specific use case from my code, in a `MethodInterceptor`:
```java
@Override
public Object invoke(final MethodInvocation invocation)
{
  return txnExecutor.doInTransaction(() ->
  {
    try
    {
      return invocation.proceed();
    }
    catch (Throwable t)
    {
      throw new Exception(t);
    }
  });
}
```

Hopefully you'd agree that this would be somewhat prettier:
```java
@Override
public Object invoke(final MethodInvocation invocation)
{
  return txnExecutor.doInTransaction(Unchecked.callable(invocation::proceed));
}
```

Plus, as you mention, symmetry is nice. `Runnable`s and `Callable`s are often found nearby each other, so having one go through `Unchecked` two lines away from another going through a raw lambda (or an ugly one that catches `Throwable`, as above) is inconsistent.

lukaseder:
OK, you're right. Thanks again for the PR!

PR #284, opened 2017-02-16
Added Async helper and SameExecutorCompletionStage.

billoneil:


src/main/java/org/jooq/lambda/concurrent/Async.java
```diff
@@ -0,0 +1,20 @@
...
+
+    }
+
+    public static <U> CompletionStage<U> supplyAsync(Supplier<U> supplier, Executor executor) {
+        return SameExecutorCompletionStage.of(CompletableFuture.supplyAsync(supplier, executor), executor);
+    }
+
+    public static CompletionStage<Void> runAsync(Runnable runnable, Executor executor) {
+        return SameExecutorCompletionStage.of(CompletableFuture.runAsync(runnable, executor), executor);
+    }
```
lukaseder: These methods are great. I think we should also handle the default case, though, i.e. omitting the `Executor` argument. WDYT?

src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Executor;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+final class SameExecutorCompletionStage<T> implements CompletionStage<T> {
+    private final CompletionStage<T> delegate;
+    private final Executor defaultExecutor;
```
lukaseder: This reference will be nullable, which means that null checks may need to be made in subsequent logic

lukaseder:
Looks great, especially the tests! thank you very much!

src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
billoneil: Should we add logic to make sure it's not null or should it be allowed to be null?

src/main/java/org/jooq/lambda/concurrent/Async.java
billoneil: I'm slightly opposed to it just because I think it will encourage users to overuse the common pool.

`Async.runAsync(() -> {...}, ForkJoinPool.commonPool())` isn't that hard to write out if that is what they really want.

I think it's fine to add to match the `CompletableFuture` api but not adding it might encourage better use. Whatever you think.

billoneil:
I will try to add tests for the rest of the methods this weekend to make sure there were no typos made along the way.

src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
lukaseder: It should be allowed to be null, I think. In that case, we'd simply use the `CompletableFuture` default, which is the default `ForkJoinPool`

src/main/java/org/jooq/lambda/concurrent/Async.java
lukaseder: It depends of course.

- One one side, you're right. It's sometimes good to force people to think.
- On the other side, that's what the folks did with the good old XML APIs. You remember how hard it is to run a simple XPath query? Exactly. In many cases, I *don't* want to think too hard about all the builders, factories and what not.

I personally favour convenience here, along with the appropriate Javadoc, of course.

lukaseder:
Perfect, thank you!

src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
billoneil:
```
CompletionStage<String> stage = Async.supplyAsync(() -> "test", executorA);
stage.thenApplyAsync(fn, null);
```

In this scenario I see three options
1. We favor the last used `Executor` (executorA)
2. Anytime a null `Executor` is passed we switch to the common pool.
3. Throw an exception when the pool is null. `CompletableFuture` does this already so there would be no code change and it would follow `CompletableFuture`'s convention.

```
    public <U> CompletableFuture<U> thenApplyAsync(
        Function<? super T,? extends U> fn, Executor executor) {
        return uniApplyStage(screenExecutor(executor), fn);
    }

    ...

    static Executor screenExecutor(Executor e) {
        if (!useCommonPool && e == ForkJoinPool.commonPool())
            return asyncPool;
        if (e == null) throw new NullPointerException();
        return e;
    }
```

I prefer option 3 but not very strongly, any of them work. If we pick option 1 or 2 it just needs to be documented.

src/main/java/org/jooq/lambda/concurrent/Async.java
billoneil: Ok sounds good.
Would you prefer this explicit style?
```
public static CompletionStage<Void> runAsync(Runnable runnable) {
    Executor executor = ForkJoinPool.commonPool();
    return SameExecutorCompletionStage.of(CompletableFuture.runAsync(runnable, executor), executor);
}
```

Or
```
public static CompletionStage<Void> runAsync(Runnable runnable) {
    return SameExecutorCompletionStage.of(CompletableFuture.runAsync(runnable), null);
}
```
This would make the decision of how to handle  null `Executor` a little more important.

src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
lukaseder: Hmm, I hadn't thought of allowing `null` in the public API... But there could be `Async.supplyAsync(() -> "test"))` which would internally set the `Executor` to `null` and then not specify any executor when delegating to the `delegate`...

I'm not sure if we should allow `null` in the public API, so I'd also prefer 3)

src/main/java/org/jooq/lambda/concurrent/Async.java
lukaseder: I'd prefer the latter because explicitly accessing `ForkJoinPool.commonPool()` would bypass some more esoteric SPI logic that is available in `CompletableFuture`:

```java
    private static final boolean useCommonPool =
        (ForkJoinPool.getCommonPoolParallelism() > 1);

    /**
     * Default executor -- ForkJoinPool.commonPool() unless it cannot
     * support parallelism.
     */
    private static final Executor asyncPool = useCommonPool ?
        ForkJoinPool.commonPool() : new ThreadPerTaskExecutor();
```

lukaseder: ... in particular, that SPI logic might change in future JDK versions, and I'd like such change to apply automatically to `Async`

billoneil: Good point.

PR #285, opened 2017-02-18
Added more SameExecutorCompletionStage tests.

billoneil:


src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
```diff
@@ -28,6 +28,9 @@
 
     @Override
     public final <U> CompletionStage<U> thenApplyAsync(Function<? super T, ? extends U> fn) {
+        if (defaultExecutor == null) {
+            return of(delegate.thenApplyAsync(fn), null);
+        }
```
billoneil: Was this what you had in mind? I think this should work. We only need to use this logic on the methods where we were supplying the default `Executor`

src/test/java/org/jooq/lambda/concurrent/SameExecutorCompletionStageTest.java
```diff
@@ -70,8 +73,228 @@ public void testThenAccept() throws InterruptedException {
...
+        assertHits(executorA, 2);
+        assertHits(executorB, 1);
+    }
+    
+    @Test
+    public void testExceptionally() throws InterruptedException {
+        CompletionStage<Void> firstCompletionStage = Async.supplyAsync(latchSupplier(latch1), executorA);
+        firstCompletionStage = firstCompletionStage.exceptionally(ex -> null);
+        latch1.await();
+        assertHits(executorA, 1);
```
billoneil: I'm not sure if this test is doing anything useful.

src/main/java/org/jooq/lambda/concurrent/SameExecutorCompletionStage.java
lukaseder: Yes, that's what I had in mind

lukaseder:
Looks good to me, thank you very much, again!

Issue #287, opened 2017-02-18
Move Async to org.jooq.lambda package


lukaseder:
@billoneil I've noticed only now that you've created a new subpackage `concurrent`. I wonder if we'll need that. I somewhat regret the existing packages. Would you agree to moving this to `org.jooq.lambda`?

----

API was not published yet prior to 0.9.13, so this isn't a breaking change.

billoneil:
Sure whatever package you prefer.

billoneil:
Should all the classes move?

PR #288, opened 2017-02-19
Moved Async and SameExecutorCompletionStage to org.jooq.lambda

billoneil:
resolves https://github.com/jOOQ/jOOL/issues/287

### DIFF
--- a/src/main/java/org/jooq/lambda/Async.java
+++ b/src/main/java/org/jooq/lambda/Async.java
@@ -1,4 +1,4 @@
-package org.jooq.lambda.concurrent;
+package org.jooq.lambda;
 
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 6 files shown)


lukaseder:
Thanks a lot!

lukaseder:
Hmm, looks like some tests fail nondeterministically: 
https://travis-ci.org/jOOQ/jOOL/builds/203172319#L896

They didn't fail on my machine

billoneil:
Looking into it now.

billoneil:
I think its possible for the `countdown` to complete

It looks like its possible to return before the count is incremented here. I can add a join before every assert to assure the `CompletionStage` is finished executing.
```java
counter.accumulateAndGet(1, (left, right) -> {
    delegate.execute(command);
    return left + right;
});
```

Issue #290, opened 2017-03-03
Joins with itself


Spike2050:
Why isn't here a shortcutmethod of being able to Join a Sequence with itself, like this (from Seq.java):

    /**
     * Cross join stream with itself
     * <p>
     * <code><pre>
     * // (tuple(1, 1), tuple(1, 2), tuple(2, 1), tuple(2, 2))
     * Seq.of(1, 2).crossJoin(Seq.of("a", "b"))
     * </pre></code>
     */
    default <U> Seq<Tuple2<U, U>> crossJoin() {
        List<U> list = toList();
        return Seq.crossJoin(Seq(list), Seq(list));
    }

Or why wouldn't that make sense? Ok, Streams are one-shot, therefore I have to convert it to a list, which isnt lazy, but it would safe me a lot of hastle writing the additional code, e.g. writing the conversion myself beforehand.

Same of course with Inner-, LeftOuter-, and RightOuterJoin.

lukaseder:
Hmm, we could support [cartesian powers](https://en.wikipedia.org/wiki/Cartesian_product#Cartesian_power) up to a certain number of degrees (although, they'd be very quickly impractical...)

What would be the point of doing this with outer joins?

Spike2050:
Hey lukaseder
I have this example for an OuterJoin (in Pseudocode):

If you have a Person(Id, ParentId) and this Dataset:
```
Lisa(1,0)
Fred(2,1)
Cindy(3,1)
```
you could make a LeftOuterJoin with d1.ParentId == d2.Id to get the Persons and their Parent in a tuple and the Persons that dont have a parent (maybe to handle these seperatly):
```
Tuple<Lisa,Null>
Tuple<Fred,Lisa>
Tuple<Cindy,Lisa>
```
About your number of degrees question, you could start with 2, I think that would cover most use cases, If you wanted more, ýou could go the established route by giving the dataset again and again via parameter. If you'd make more degress, I think you would also need a TriPredicate, QuadPredicate, etc?

Your thoughts?

lukaseder:
OK, I see. That might make sense indeed, at least for degree 2. I don't think it would be a reasonable extension for higher degree joins as the ordering semantics might become quite complex.

Spike2050:
Yes, so I suggest, that a join (l/r/i/c) without another stream as parameter would be automatically a Self-Join, like this:

```
default <T> Seq<Tuple2<T, T>> crossJoin()
[...]

PR #292, opened 2017-03-05
Add support for self (left / right / inner / cross) join

Spike2050:
Changes as discussed in issue #290.

Has (passing) tests and added description in readme.md.

I also adjusted some comments from the other joins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,8 @@ Seq.of(1, 2, 3, 4).grouped(i -> i % 2);
 // (tuple(1, 1), tuple(2, 2))
 Seq.of(1, 2, 4).innerJoin(Seq.of(1, 2, 3), (a, b) -> a == b);
 
-// (tuple(1, 1), tuple(2, 2))
-Seq.of(1, 2).innerSelfJoin((t, u) -> t == u)
+// (tuple(1, 2), tuple(2, 1))
+Seq.of(1, 2).innerSelfJoin((t, u) -> t != u)
 
 // (1, 0, 2, 0, 3, 0, 4)
 Seq.of(1, 2, 3, 4).intersperse(0);
@@ -84,7 +84,7 @@ Seq.of(1, 2, 3).join("|", "^", "$");
 Seq.of(1, 2, 4).leftOuterJoin(Seq.of(1, 2, 3), (a, b) -> a == b);
 
 // (tuple(tuple(1, 0), NULL), tuple(tuple(2, 1), tuple(1, 0)))
-Seq.of(new Tuple2<Integer, Integer>(1, 0), new Tuple2<Integer, Integer>(2, 1)).leftOuterSelfJoin((t, u) -> t.v2 == u.v1)
+Seq.of(tuple(1, 0), tuple(2, 1)).leftOuterSelfJoin((t, u) -> t.v2 == u.v1)
 
 // (1, 2)
 Seq.of(1, 2, 3, 4, 5).limitWhile(i -> i < 3);
@@ -99,7 +99,7 @@ Seq.of(new Object(), 1, "B", 2L).ofType(Number.class);
 Seq.of(1, 2, 4).rightOuterJoin(Seq.of(1, 2, 3), (a, b) -> a == b);
 
 // (tuple(NULL, tuple(1, 0)), tuple(tuple(1, 0), tuple(2, 1)))
-Seq.of(new Tuple2<Integer, Integer>(1, 0), new Tuple2<Integer, Integer>(2, 1)).rightOuterSelfJoin((t, u) -> t.v2 == u.v1)
+Seq.of(tuple(1, 0), tuple(2, 1)).rightOuterSelfJoin((t, u) -> t.v2 == u.v1)
 
 // tuple((1, 3), (2, 4))
 Seq.of(1, 2, 3, 4).partition(i -> i % 2 != 0);
--- a/README.md
+++ b/README.md
@@ -41,6 +41,9 @@ Seq.of(1, 2, 3, 4).containsAny(2, 5);
 // (tuple(1, "A"), tuple(1, "B"), tuple(2, "A"), tuple(2, "B"))
 Seq.of(1, 2).crossJoin(Seq.of("A", "B"));
 
+// (tuple(1, 1), tuple(1, 2), tuple(2, 1), tuple(2, 2))
+Seq.of(1, 2).crossSelfJoin()
+
 // (1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, ...)
 Seq.of(1, 2, 3).cycle();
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -62,6 +65,9 @@ Seq.of(1, 2, 3, 4).grouped(i -> i % 2);
 // (tuple(1, 1), tuple(2, 2))
 Seq.of(1, 2, 4).innerJoin(Seq.of(1, 2, 3), (a, b) -> a == b);
 
+// (tuple(1, 1), tuple(2, 2))
+Seq.of(1, 2).innerSelfJoin((t, u) -> t == u)
```
lukaseder: Hmm, this would be a rather inefficient alternative to:

```java
Seq.of(1, 2).map(i -> tuple(i, i))
```

Perhaps, a more convincing example might be interesting here, similar to your outer join examples, where child / parent relationships are joined. What do you think?

README.md
```diff
@@ -77,6 +83,9 @@ Seq.of(1, 2, 3).join("|", "^", "$");
 // (tuple(1, 1), tuple(2, 2), tuple(4, null))
 Seq.of(1, 2, 4).leftOuterJoin(Seq.of(1, 2, 3), (a, b) -> a == b);
 
+// (tuple(tuple(1, 0), NULL), tuple(tuple(2, 1), tuple(1, 0)))
+Seq.of(new Tuple2<Integer, Integer>(1, 0), new Tuple2<Integer, Integer>(2, 1)).leftOuterSelfJoin((t, u) -> t.v2 == u.v1)
```
lukaseder: For brevity's sake, we can replace the constructor call with the `Tuple.tuple()` convenience method and assume a static import, so:

```java
Seq.of(tuple(1, 0), tuple(2, 1))...
```

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -203,11 +203,27 @@
...
+        return Seq.crossJoin(seq(list), seq(list));
+    }
+
+    /**
      * Inner join 2 streams into one.
      * <p>
      * <code><pre>
      * // (tuple(1, 1), tuple(2, 2))
-     * Seq.of(1, 2, 3).innerJoin(Seq.of(1, 2), t -> Objects.equals(t.v1, t.v2))
+     * Seq.of(1, 2, 3).innerJoin(Seq.of(1, 2), (t, u) -> Objects.equals(t, u))
```
lukaseder: Thanks for these. Would you mind fixing these in a separate pull request, as they're not strictly related to this particular change...

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -231,26 +247,43 @@
...
+     */
+    default Seq<Tuple2<T, T>> innerSelfJoin(BiPredicate<? super T, ? super T> predicate) {
+
+        // This algorithm isn't lazy and has substantial complexity for large argument streams!
+        List<? extends T> list = toList();
+
+        return seq(list).flatMap(t -> seq(list)
+                                     .filter(u -> predicate.test(t, u))
+                                     .map(u -> tuple(t, u)));
+    }
```
lukaseder: Hmm, I wonder if we perhaps would be better off if this method delegates to the existing `innerJoin()` method, and we start optimising these joins in a different task...

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -283,18 +316,37 @@
...
         return flatMap(t -> seq(list)
-              .filter(u -> predicate.test(t, u))
-              .onEmpty(null)
-              .map(u -> tuple(t, u)))
-              .onClose(other::close);
+                           .filter(u -> predicate.test(t, u))
+                           .onEmpty(null)
+                           .map(u -> tuple(t, u)))
+                           .onClose(other::close);
+    }
```
lukaseder: Thanks for fixing the formatting. Perhaps, this could be done in a separate task as well. In particular, now, the `onClose()` method is not aligned correctly according to your fix's intention.

lukaseder: Thanks for fixing the formatting. Perhaps, this could be done in a separate task as well. In particular, now, the `onClose()` method is not aligned correctly according to your fix's intention.

README.md
Spike2050: Ill just change that to 

// (tuple(1, 2), tuple(2, 1))
Seq.of(1, 2).innerSelfJoin((t, u) -> t != u)

README.md
Spike2050: np

src/main/java/org/jooq/lambda/Seq.java
Spike2050: Well I could just do this

```
default Seq<Tuple2<T, T>> innerSelfJoin(BiPredicate<? super T, ? super T> predicate) {

    // This algorithm isn't lazy and has substantial complexity for large argument streams!
    List<? extends T> list = toList();

    return seq(list).innerJoin(seq(list), predicate);
}
```

Spike2050:
I made another pull request with just the comments

README.md
lukaseder: Yeah, nice thinking. That's a common use-case in SQL too

src/main/java/org/jooq/lambda/Seq.java
lukaseder: Even better

Spike2050:
I incorporated the discussed changes, you can check it again.

lukaseder:
That looks great, thank you very much again for your contribution!

Spike2050:
Pleasure

When do you plan your next release?

lukaseder:
I guess we have enough new things. Will look into it these days

Spike2050:
Ever thought about doing an official Snapshot/Nightly/Alpha?

I mean I did install a private one on my nexus, to test the selfjoins with my other projects, but an official one would be nice.

lukaseder:
If there's something OOTB that does this on Github with absolutely no manual work involved, sure. Do you know any?

Spike2050:
Well I work with Jenkins, which does this for me, but I think Travis can do that too.

I found this [link ](https://coderwall.com/p/9b_lfq/deploying-maven-artifacts-from-travis) which descripes a Travis Setup to deploy to Maven, including SNAPSHOT-version. But as I see it you might need to hide your Travis configuration from public Github repo, because it would store your Maven repo login

billoneil:
I don't know if its worth publishing snapshots to maven. If you really want
to allow users to use unreleased versions, a very little effort solution
could be jitpack.

You grant them access to the repo. It clones the repo and builds the code
from a tag / branch / git hash in a docker container on the fly and serves
the produced Jar. It works with maven and gradle if you add them as a
repository.

https://jitpack.io/docs/#building-with-jitpack

lukaseder:
Hmm, yes, I could use the jOOQ CI server (Jenkins) to deploy nightlies to Maven Central... But that's already too much effort, given the relatively little progress this library makes between releases :)

And Jitpack sounds nice, but another infrastructure dependency to look after.

I'll wait with this, until more users request it (also for the other jOO* libraries).

Spike2050:
Hey lukaseder.

What ever came of the realease you were looking into?

Greets

lukaseder:
I'm sorry - not sure if I understand, what are you asking me, @Spike2050 ?

bendem:
> When do you plan your next release?

> I guess we have enough new things. Will look into it these days

I think he's speaking about the fact that the last release you did was more than a year ago and that there are a [bunch of new stuff to release now](https://github.com/jOOQ/jOOL/compare/429e850...aab32d1) (I'd totally be interested in a new release as well) :).

Spike2050:
> I think he's speaking about the fact that the last release you did was more than a year ago and that there are a bunch of new stuff to release now (I'd totally be interested in a new release as well) :).

Well that :-)

PR #293, opened 2017-03-05
Ident Formatting in Joins and fixed comments

Spike2050:
Hi lukaseder

Here are the comment and (fixed) identation adjustments from my other pull request, If you accept this I'll put my other changes on top of that.


lukaseder:
Thank you very much, and again: Nice catch!

Issue #295, opened 2017-03-11
Throwing Excptions via Optionals?


Spike2050:
Hi lukaseder

I have a question about the sneakythrow function. You throw any Throwable with these methods

```java
static void sneakyThrow(Throwable throwable) {
    SeqUtils.<RuntimeException>sneakyThrow0(throwable);
}

@SuppressWarnings("unchecked")
static <E extends Throwable> void sneakyThrow0(Throwable throwable) throws E {
    throw (E) throwable;
}
```
I don't understand how you can do that, how can you cast any Exception to a RuntimeException, without the compiler complaining?

You could use this to throw Exceptions inside a sequence lambda if you add a function to Seq.java like this:
```java
default <X extends Throwable> Seq<T> onOptionalThrow(Function<? super T, Optional<? extends X>> function) {
    return peek(e -> {
        Optional<? extends X> opt = function.apply(e);
        if (opt.isPresent()) {
            sneakyThrow(opt.get());
        }
    });
}
```
You could use this function to do stuff like this:
```java
Seq.of(1, 2, 3).onOptionalThrow(t -> {
     if (t > 3) {
        return Optional.of(new Exception());
    }
    return Optional.empty();
}).toList();
```
I wrote a test and it seems to work, but I can't confidently say why. Why is this a bad idea or what have i forgotten.

lukaseder:
Yeah, that's magic, isn't it? The `RuntimeException` generic type binding does not result in a type cast, so there's no class cast exception. However, it does trick the compiler into believing that the exception being thrown is unchecked.

Hope this helps

Spike2050:
Well wait :-)

Are you interested to add this function into the library:
```java
default <X extends Throwable> Seq<T> onOptionalThrow(Function<? super T, Optional<? extends X>> function) {
    return peek(e -> {
        Optional<? extends X> opt = function.apply(e);
        if (opt.isPresent()) {
            sneakyThrow(opt.get());
        }
    });
}
```
You could throw exceptions without converting them to RunitmeException inside a long Sequence chain and without a surrounding try-catch-block.

lukaseder:
Aha, I see. Sorry, I misunderstood - thought you were just showing an example. I currently don't see this functionality as interesting enough to add to the `Seq` API.  Your suggested example:

```java
Seq.of(1, 2, 3).onOptionalThrow(t -> {
     if (t > 3) {
        return Optional.of(new Exception());
    }
    return Optional.empty();
}).toList();
```

[...]

PR #297, opened 2017-03-12
Add onOptionalThrow for throwing Exceptions via lambda return.

Spike2050:
Enables throwing of Exception without conversion into RuntimeException and try/catch-block. Like this

```java
Seq.of(1, 2, 3).onOptionalThrow(t -> {
     if (t > 3) {
        return Optional.of(new Exception());
    }
    return Optional.empty();
}).toList();
```
Your thoughts?


lukaseder:
Please, let's first discuss this in https://github.com/jOOQ/jOOL/issues/295. I doubt this will make it into the library in its current form.

Issue #91, opened 2015-06-01
Add Seq.peekThrowable(Consumer<Throwable>)


lukaseder:
There is currently no fluent way of handling exceptions in a `Stream`. While `Stream` extends `AutoCloseable` and provides an `onClose(Runnable)` hook for actions that are to be taken before the actual closing, there is no such thing for exceptions.

We should add an `peekThrowable(Consumer<Throwable>)` hook that allows `Consumers` to be added to consume all sorts of exceptions that may occur.

The actual API is not yet fully defined and may differ, e.g. by taking inspiration from the `Try` monad

tkroman:
@lukaseder, do you have any thoughts on this? I believe, `CompletableFuture`'s API is a rather well designed one, although a little bloated with (maybe) too much methods responsible for handling the computation result, so maybe this feature's implementation could borrow something from there?

lukaseder:
I haven't thought all situations through yet. I'm afraid that it might not be a very good idea to implement such a method in a `Stream`. Perhaps, there are situations where exceptions will escape...

lukaseder:
An implementation suggestion was made by @tlinkowski in https://github.com/jOOQ/jOOL/pull/299

lukaseder:
Implemented through https://github.com/jOOQ/jOOL/pull/315

PR #299, opened 2017-04-24
Add Seq.onThrowable() and Seq.onThrowableAs()

tlinkowski:
I provide two `Seq` methods for consideration:
- `onThrowable(Consumer<? super Throwable> throwableHandler)`
- `onThrowableAs(Function<? super Throwable, ? extends Throwable> throwableWrapper)`

The naming and implementation is subject to improvements.

Rationale:
- I have a method that returns a `Seq` which may then be consumed anywhere else
- the consumption of the `Seq` might throw some recoverable exceptions
- I needed to wrap these exceptions (that might be thrown deep down in the call stack) with a "parent" exception (that can be created only higher up the call stack)

PS. I haven't provided the unit tests.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -440,6 +440,20 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
     }
 
     /**
+     * If a throwable occurs, call the <code>throwableHandler</code> on it before rethrowing.
+     */
+    default Seq<T> onThrowable(Consumer<? super Throwable> throwableHandler) {
+        return SeqUtils.transform(this, (delegate, action) -> {
+            try {
+                return delegate.tryAdvance(action);
+            } catch (Throwable throwable) {
+                throwableHandler.accept(throwable);
+                throw throwable;
+            }
+        });
+    }
+
+    /**
      * Concatenate two streams.
      * <p>
      * <code><pre>
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -454,6 +454,13 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
     }
 
     /**
+     * If a throwable occurs, wrap it in a throwable from the <code>throwableWrapper</code>.
+     */
+    default Seq<T> onThrowableAs(Function<? super Throwable, ? extends Throwable> throwableWrapper) {
+        return onThrowable(throwable -> SeqUtils.sneakyThrow(throwableWrapper.apply(throwable)));
+    }
+
+    /**
      * Concatenate two streams.
      * <p>
      * <code><pre>


lukaseder:
Hmm, that's an interesting idea. I'd rename this to `peekThrowable()` as it has the same semantics as the existing `peek()` method.

Now about that throwable wrapper method: I feel a bit uneasy about it. Are you sure this will correctly work in all cases of streams? I would definitely not want to merge this without extensive unit testing...

tlinkowski:
`peekThrowable` sounds good.

As for the wrapper method - I have no idea :) I imagine this nesting can go very deep so perhaps the API and the implementation should be modified to prevent/limit this, e.g. with: `<X extends Throwable> Seq<T> onThrowableAs(Class<X> throwableType, Function<X, ? extends Throwable> throwableWrapper)`

I did use this `onThrowableAs` with a certain tree structure (with nesting reaching up to 20-30 levels) and I haven't experienced any problems but I limited the wrapping to one exception only:
- I used a modifiable "collector" exception that stored the original cause and "collected" data from all the levels
- then, at the top, I caught this "collector" exception and used its data and its cause to create another, "final" exception that I threw
 
Also note, that right now `onThrowableAs` is implemented using `onThrowable/peekThrowable` so theoretically they might suffer from the same problems.

Anyway, I agree that extensive unit testing is highly recommended.

tlinkowski:
BTW. I just found #91 of which `onThrowable` is a duplicate.

lukaseder:
OK, thanks for digging up the duplicate (https://github.com/jOOQ/jOOL/issues/91). So, I'll close this issue here.

I'm still not 100% convinced that this is feasible / useful, though. Perhaps if I see a concrete example in code (I mean the client code, not just the `Seq` implementation)...

tlinkowski:
I understand though I'm a bit confused why you close both #299 and #91. I kind of expected one of them could stay open so that anyone else can potentially back this up.

It seems to me this feature (at least `peekThrowable/onThrowable/onFailure`) is really useful because right now `Seq` has absolutely no way of handling `Exception`s outside of `Seq` consumption site.

lukaseder:
Egh, sorry, that was an accident (closing #91)

> It seems to me this feature (at least onThrowable/peekThrowable - but onThrowableAs/wrapThrowable also) is really useful because right now Seq has absolutely no way of handling Exceptions outside of Seq consumption site.

I know, I wish there was such a thing, but I have been looking into this when trying to implement a stream completion hook: http://stackoverflow.com/q/34753078/521799. It appears that

- This was out of scope for the Java 8 EGs
- It is actually rather complicated to achieve, i.e. to be sure that any such logic is truly executed

So, as I said, I'm currently reluctant to investigate this again myself, as I doubt this feature can actually be implemented. I'm happy to be proven wrong :)

tlinkowski:
Well, I guess I'm not competent enough to provide a proof in the true meaning of this word.

But here is my reasoning:
- **every** `Stream`/`Seq` operation is `Spliterator`-based - in other words, `Spliterator` is the ultimate source of the `Stream` content
- even `Iterator` returned by terminal `iterator()` is `Spliterator`-based (see `Spliterators$Adapter`)
- the only way in which `Spliterator` traverses elements (i.e. "provides data") is by `tryAdvance` (there is also `trySplit` but in `SeqUtils` it's a no-op)
- as a result, if you wrap calls to `tryAdvance` at some level with a `try-catch` block (or any other block, for that matter), this block -  according to my best knowledge - **must** be executed at all times

As for the StackOverflow issue (`onComplete`) you pointed me to - it's very interesting, and I don't know how to solve the problem raised there. But I'm pretty confident that `onComplete` and `onThrowable` are two really different things because:
- `onComplete` looks for the **end** of the `Stream` consumption, and as @amaembo pointed out on StackOverflow, `Spliterator` can just be abandoned in the middle of traversal so there's **no way** for you to determine if `Stream` consumption completed or not
- `onThrowable`, on the other hand, does not look for anything - it simply acts as a **traversal observer**, i.e. an observer of `tryAdvance` calls

This brings us, however, to an important remark (which I didn't think of before) - `onThrowable` can handle **only** errors occurring during the **traversal** of the source. For example, if you call any reduction terminal operation (like `reduce` or `collect`), it can't handle accumulation nor combining errors. So I guess its name should be adapted to better reflect its purpose, i.e. `onTraversalThrowable` or `peekTraversalThrowable`.

To sum up, it's quite possible I'm still missing something important but right now I still feel that this would make a valuable addition to `Seq`. On the other hand, I'd understand if you preferred to add a method like that only if it supported absolutely all (and not only traversal-related) cases.

lukaseder:
Thanks for the analysis. I agree with you mostly. I've renamed https://github.com/jOOQ/jOOL/issues/91 to be called `peekThrowable`. I don't think we need any new names. Just like the `peek()` method, `peekThrowable()` should be only able to see what's being throw *upstream*. For example:

```java
Seq.seq(source)
   .filter(op1)
   .filter(op2)
   .peekThrowable(e -> ... /* only exceptions from op1 and op2 */
   .filter(op3)
   .peekThrowable(e -> ... /* exceptions from op1, op3, or op3 */
   .collect(collector);
```

As you can see, I don't think a user would expect such a `peekThrowable()` method to see anything that's going wrong inside of the `collector`.

Having said so, I think we could implement #91, with `peekThrowable()` semantics.

Issue #179, opened 2016-01-11
Add Seq<T> onEmptySwitch(Supplier<? extends Stream<? extends T>>)


lukaseder:
As suggested by @johnmcclean-aol (https://github.com/jOOQ/jOOL/issues/177#issuecomment-170689447), an `onEmptySwitch(Supplier<? extends Stream<? extends T>>)` method might be useful in some situations.

Criticism:

- The current `onEmptyXXX()` names were inspired by `Optional.orElseXXX()`. We should think about the term `switch`, or if there is any more suitable term for this action
- Due to generic type erasure, we cannot overload a single `onEmptySwitch(Supplier<E>)` method with different bounds for `<E>`. What would be the best upper bound? `Seq`? `Stream`? `Iterable`?

tlinkowski:
Sample implementation:
> default Seq<T> onEmptySwitch(Supplier<? extends Stream<T>> supplier) {
    Supplier<? extends Stream<T>> lazy = () -> {
        Iterator<T> it = iterator();
        return it.hasNext() ? seq(it) : supplier.get();
    };
    return of(lazy).flatMap(Supplier::get);
}

As for your question about overloads, maybe different names like: `onEmptyGetStream`, `onEmptyGetSeq`, and `onEmptyGetIterable`?

lukaseder:
Thank you for your suggestion, @tlinkowski. Hmm, I'm not sure it is correct, though. Why are you flatmapping your "lazy" stream with `Supplier` get? If the `this` stream contains 3 elements, for instance, that would call `Supplier.get()` 3 times.

But still, the fact that we'll need these wonky overloads to be sure we can get around type erasure, I'm still a bit hesitant to add this at all.

tlinkowski:
> Why are you flatmapping your "lazy" stream with Supplier get? If the this stream contains 3 elements, for instance, that would call Supplier.get() 3 times.

No, no - this `Supplier::get` is called only on the `lazy` variable, and not on the given `supplier`. So it's always called only once because there is only one `lazy` variable. But perhaphs I shouldn't have provided this implementation in such form, and I should have provided it using the proposed `Seq.lazy` [#302]:

```java
default Seq<T> onEmptySwitch(Supplier<? extends Stream<T>> supplier) {
    return lazy(() -> {
        Iterator<T> it = iterator();
        return it.hasNext() ? seq(it) : supplier.get();
    });
}
```

lukaseder:
[...]

PR #300, opened 2017-04-24
Add Seq.lazy() and Seq.onEmptySupply()

tlinkowski:
Sorry that there are several things at once here but I didn't feel like creating so many PRs (so just pick whatever you like).

Three commits:
1. Some documentation improvements for `Seq.onEmpty*` methods
2. `static Seq.lazy(Supplier<? extends Seq<T>>)` method that returns a lazy `Seq` from the given supplier
(useful to return single-valued sequences as lazy ones, e.g. `Seq.lazy(() -> Seq.of(computeValue()))`)
3. `Seq.onEmptySupply(Supplier<? extends Seq<T>>)` which allows to return an entire `Seq` if the original `Seq` is empty (in contrary to `onEmptyGet` which allows to return only a single-valued `Seq`)
(the `Iterator`-based implementation may be non-optimal but I'm not sure how to do it better)

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -390,7 +390,7 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
     }
 
     /**
-     * Produce this stream, or an alternative stream from the
+     * Produce this stream, or an alternative stream with the
      * <code>value</code>, in case this stream is empty.
      */
     default Seq<T> onEmpty(T value) {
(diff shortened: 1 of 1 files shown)


lukaseder:
Thank you very much for your suggestion.

> Sorry that there are several things at once here but I didn't feel like creating so many PRs (so just pick whatever you like).

I'm sorry I cannot merge this. I'd like to discuss each individual idea first (and individually to avoid a messy discussion).

Thanks
Lukas

tlinkowski:
Sure, I understand.

Only now I found out that my 'Seq.onEmptySupply' duplicates #179. So perhaps it's enough if I leave my implementation proposal there as a comment.

However, would you like my to create two separate PRs - one for documentation fixes, and the other for `Seq.lazy`?

lukaseder:
> So perhaps it's enough if I leave my implementation proposal there as a comment.

Yes, that's perfect, thank you!

> However, would you like my to create two separate PRs - one for documentation fixes, and the other for Seq.lazy?

That would be great, yes. Note, if in doubt (and to avoid extra work on your side), you can also just create an issue describing the API (e.g. `Seq.lazy()`). That might be less effort than an actual PR. OTOH, the PR for the documentation fixes can be done with GitHub's web UI directly

PR #302, opened 2017-04-24
Add Seq.seq(Supplier<? extends T>)

tlinkowski:
`Seq.lazy(Supplier<? extends Stream<T>>)` would take a `Stream` supplier and turn it (lazily) into a `Seq`.

If we wanted to support other types (e.g. `IntStream`, `Iterable<T>`, `Optional<T>`, `T`) we'd have to use different names for each method because of `Supplier`'s type erasure, e.g. `lazySeq`, `lazyIntSeq`, `lazyIterable`, `lazyOptional`, `lazyValue` (they would generally correspond to what can be non-lazily obtained by `Seq.of` and `Seq.seq`).

For me, however, simple `Seq.lazy()` would be quite enough because it's just a matter of calling `Seq.of` or `Seq.seq` inside the lazy `Supplier` body.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -4621,7 +4621,14 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
     static <T> Seq<T> generate(Supplier<? extends T> s) {
         return seq(Stream.generate(s));
     }
-    
+
+    /**
+     * Lazily produce a <code>Seq</code>.
+     */
+    static <T> Seq<T> lazy(Supplier<? extends Stream<T>> s) {
+        return seq(Stream.of(s).flatMap(Supplier::get));
+    }
+
     /**
      * Wrap an array slice into a <code>Seq</code>.
      * 


lukaseder:
Thank you very much for your suggestion. Would you mind showing 1-2 use-cases for `Seq.lazy()`? Given that a `Stream` itself is already a lazy data structure (it might not be because of some implementation flaws, but that can be fixed), I'm not sure what the value of this method will be.

tlinkowski:
Sure.

1. My real use case: assume you have a `computeExpensiveValue()` method. You want to return its return value as a single-valued `Seq`. This `Seq` is appended to some other `Seq` somewhere further up the call stack, filtered, and generally undergoes many operations. Finally, you just call `findFirst` on the top-level `Seq`. So you don't know if the `Seq` consumption will ever reach `computeExpensiveValue()` or not. And if it doesn't, and you have many such `computeExpensiveValue()` functions in the resulting `Seq`, it hinders performance.
2. [#179] `Seq.onEmptySwitch`: my implementation requires a call to `Stream.iterator` and then `Iterator.hasNext` which actually computes the first element (`Spliterator$Adapter`). Wrapping this operation inside `Seq.lazy` guarantees the returned `Seq` is truly lazy.
3. [#199] `Seq.shuffle`: you could use it to implement it in an easier way (it's already lazy).
4. [#195] `Seq.reverse`, `Seq.window`, and several more: all you'd need to do to make them lazy is to wrap the current code with `return Seq.lazy(() -> { /*current code*/});`).
5. [#195] `Seq.*Join`: these would not be fully lazy with `Seq.lazy` (their laziness would "work" only until the first element was requested, and by "fully" I mean not calling `toList()`). Such simple laziness would still be better than what we have now, i.e. instant consumption of the entire `Stream`. But I think there's a much better way to do it (details in #195).

lukaseder:
Thanks for your ideas. OK, I see. I suspect I was a bit troubled by your flatmapping implementation, rather than looking at the API suggestion. In fact, the simplest implementation I can see is:

```java
Seq.generate(supplier).limit(1)
```

And if we rename the method, e.g. to `Seq.seq(Supplier)`, then we have something conceptually similar to `Seq.seq(Optional)`.

lukaseder:
I'll merge and adapt this.

tlinkowski:
Thanks for merging this. I must say that your new `seq` methods are very nice and useful but they are actually different from what I proposed :)
- I proposed a way to create a `Seq` from a `Supplier<Seq<T>>` - essentially lazy-constructing an arbitrary `Seq`
- you implemented a way to create a `Seq` from a `Supplier<T>` - also lazy but limited to a single element

And I must admit I like this name `lazy` because it best expresses the API intention. The implemention, however, and as you experienced yourself, is quite non-intuitive and when applied directly in code is hardly readable. Fortunately, name `seq` with `Supplier` argument has just been taken by you so maybe `lazy` will make it :) If you decide to include this at all, of course.

lukaseder:
I shouldn't have renamed this issue but created a new one, sorry.

You're right, the two ideas are not the same. I'm still struggling with seeing the usefulness of this API despite your examples. I rest my case, most examples you're mentioning are poor implementations (eager allocation of a buffer with `toList()`) and those should be improved, not wrapped.

tlinkowski:
It just came to my mind - is it a good idea to have overloads for a method taking lambdas as parameters?

Maybe instead `seq(Supplier)` and `seq(IntSupplier)` we should have the following names?
- `supply(Supplier<? extends T>)`
- `supplyInt(IntSupplier)`
- `supplyLong(LongSupplier)`
- `supplyDouble(DoubleSupplier)`

And even an extra method like:
- `supplyOptional(Supplier<Optional<? extends T>>)` [#306]

lukaseder:
Sorry, I'm a bit confused (really shouldn't have renamed the issue). I'm not sure if we really need these at the moment, as we don't have an `IntSeq`, `LongSeq`, etc. so boxing happens anyway for the first 4.

Not sure about `supplyOptional()` either...

tlinkowski:
Sorry, maybe I wasn't specific enough. You yourself added the following methods to `Seq`:
```java
static <T> Seq<T> seq(Supplier<? extends T> s);
static Seq<Integer> seq(IntSupplier s);
static Seq<Long> seq(LongSupplier s) ;
static Seq<Double> seq(DoubleSupplier s);
```

So there are two aspects:
1. One that I did not address - whether these overloads for primitives are needed at all because the primitives are boxed anyway (e.g. you don't have such overloads for `Seq.generate`).
2. One that I did address - whether these overloads are not susceptible to some compilation problems, as you warned in your [blog post](https://blog.jooq.org/2015/01/29/you-will-regret-applying-overloading-with-lambdas/). They compile fine in the tests so I'm not sure if this really is a problem but that's why I proposed to change their names.

As for `supplyOptional` - you're not sure about the name, or about the idea? The idea consists in lazily constructing a `Seq` from an `Optional` instead of writing:
```java
Seq.seq(() -> computeOptional()).filter(Optional::isPresent).map(Optional::get)
```

lukaseder:
Oh, I see. You're right. Perhaps I've "accidentally" hit a safe spot with the overloading of methods taking lambdas with primitives / non-primitives. But let's stay on the safe side. I'll remove the overloads. They're useful only if someone passes a non-lambda to the method.

About `supplyOptional`: I'm not sure what a `Supplier<Optional<T>>` is. It doesn't seem to be a very "interesting" type. Too edge-casey. I'd rather not add this to the library.

tlinkowski:
Well, I understand your objections when looking at `Supplier<Optional<T>>` on its own. But please note that:
- `Seq.of(T)` is to `Seq.seq(Supplier<T>)` 
- what `Seq.seq(Optional<T>)` would be to `Seq.supplyOptional(Supplier<Optional<T>>)`

The naming, however, is very confusing so I take back `supplyOptional`'s name - it only adds to the confusion.

But please let me justify some more - we now have (more or less) the following three categories:
1. `Seq.of` - trivial direct `Seq` creation
2. `Seq.seq` - trivial indirect `Seq` creation
3. `Seq.iterate`, `Seq.generate`, `Seq.cycle`, etc. - non-trivial `Seq` creation

And these categories are very nice but they have one flaw - they don't take laziness into account, which is:
- trivial + non-lazy:
`Seq.of(T)` [A]
`Seq.of(T...)` [B]
`Seq.seq(Optional<T>)` [C]
`Seq.seq(Optional<T>...)` [D]
`Seq.seq(T[], int, int)`
`Seq.seq(Map<T>)`, `Seq.seq(Iterable<T>)`
- trivial + non-lazy / lazy:
`Seq.seq(Iterator<T>)` (non-lazy because `Iterator` is by contract a "pointer" over a `Collection` - in practice, though, it's often a lazy source of elements)
- trivial + lazy:
`Seq.seq(Supplier<T>)` [A]
`Seq.{missing}(Supplier<T>...)` [B]
`Seq.{missing}(Supplier<Optional<T>>)` [C]
`Seq.{missing}(Supplier<Optional<T>>...)` [D]
`Seq.seq(Stream<T>)`, `Seq.seq(Spliterator<T>)`, `Seq.seq(Enumeration<T>)`
- non-trivial + lazy:
`Seq.iterate`, `Seq.generate`, `Seq.cycle`, etc.

Since we now have a lazy version of the [A] method, I'd like to have a lazy version of the [C] method too (plus [B] and [D] too, if possible). However, I don't have a clue how to name them properly.

lukaseder:
I think you're starting to overengineer this problem :)

tlinkowski:
Well, it may be so :) I just needed a way to create a `Seq` from `Optional` without computing this `Optional` beforehand (hence initial `Seq.lazy` and current `Seq.supplyOptional` proposals). Sounds quite reasonable to me, still.

A pity I wrote all this here, in a closed PR, and not in an open issue. But as I don't have the heart to pester you for this anymore I guess I'll just let it be :)

Issue #303, opened 2017-04-24
Add static Seq.iterateWhilePresent/iterateUntilAbsent


tlinkowski:
Method `static Seq<T> iterateWhilePresent(T seed, Function<? super T, Optional<? extends T>> generator)` would iterate until the generator returns an empty `Optional`.

Example, non-optimal implementation (with nullable `seed`) would be:
```java
UnaryOperator<Optional<? extends T>> f = optional -> optional.flatMap(generator::apply);
return Seq.iterate(Optional.ofNullable(seed), f).limitWhile(Optional::isPresent).map(Optional::get);
````

Rationale: if we have a tree-like structure where each node stores a nullable reference to its parent (null if no parent), the "ancestry line" of the tree can easily be found using `Seq.iterateWhilePresent(treeNode, TreeNode::findParent)` where `TreeNode.findParent()` returns `Optional<TreeNode>`.

PS. It would be equivalent to https://github.com/poetix/protonpack#unfold

lukaseder:
That's interesting. Do note there will be a similar method in Java 9's `Stream`
http://download.java.net/java/jdk9/docs/api/java/util/stream/Stream.html#iterate-T-java.util.function.Predicate-java.util.function.UnaryOperator-

I guess from an implementation perspective, we could definitely take inspiration from there. If I remember correctly, @amaembo implemented that, or at least tuned it quite nicely.

I somehow prefer the boolean type as an indicator of stream termination, as that goes well with the existing `limitWhile()` and similar methods...

What do you think?

tlinkowski:
Oh, I didn't realize Java 9 will have that - thanks for pointing that out to me! As to what I think, I think this Java 9's API is actually worse :) I mean, it's like `Iterator` where you have separate logic for `hasNext` and `next` when quite often they correspond to the same (possibly expensive) operation. Even in `Spliterator` they gave it up for a single method (`tryAdvance`). And in Guava, they have `AbtractIterator` that also has only one method (`computeNext`).

Note that if you wanted to apply this Java 9's API to the `TreeNode` example I mentioned earilier (with its `findParent()` returning `Optional`), it would be quite cumbersome (`Stream.iterate(Optional.of(treeNode), Optional::isPresent, optional -> optional.get().findParent()).map(Optional::get)`).

I'm not sure if it's good practice but right now I use `Optional` as return type almost everywhere where I previously used `@Nullable`, and that's why I much prefer this `Optional`-based API.

[...]

PR #304, opened 2017-04-26
Added Seq.iterateWhilePresent

tlinkowski:
Implementation for #303.

### DIFF
--- a/src/test/java/org/jooq/lambda/SeqTest.java
+++ b/src/test/java/org/jooq/lambda/SeqTest.java
@@ -1543,6 +1543,22 @@ public class SeqTest {
     }
 
     @Test
+    public void testIterateWhilePresent() {
+        assertEquals(
+              asList(),
+              Seq.iterateWhilePresent(null, i -> Optional.empty()).toList()
+        );
+        assertEquals(
+              asList(2),
+              Seq.iterateWhilePresent(2, i -> Optional.empty()).toList()
+        );
+        assertEquals(
+              asList(2, 4, 16, 256, 65536),
+              Seq.iterateWhilePresent(2, i -> i <= 256 ? Optional.of(i * i) : Optional.empty()).toList()
+        );
+    }
+
+    @Test
     public void testOfType() {
         assertEquals(asList(1, 2, 3), Seq.of(1, "a", 2, "b", 3, null).ofType(Integer.class).toList());
         assertEquals(asList(1, "a", 2, "b", 3), Seq.of(1, "a", 2, "b", 3, null).ofType(Serializable.class).toList());
--- a/src/test/java/org/jooq/lambda/SeqTest.java
+++ b/src/test/java/org/jooq/lambda/SeqTest.java
@@ -1544,10 +1544,7 @@ public class SeqTest {
 
     @Test
     public void testIterateWhilePresent() {
-        assertEquals(
-              asList(),
-              Seq.iterateWhilePresent(null, i -> Optional.empty()).toList()
-        );
+        assertThrows(NullPointerException.class, () -> Seq.iterateWhilePresent(null, i -> Optional.empty()).toList());
         assertEquals(
               asList(2),
               Seq.iterateWhilePresent(2, i -> Optional.empty()).toList()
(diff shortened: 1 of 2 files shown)


lukaseder:
Thank you very much for your suggestion. I'll review shortly

src/test/java/org/jooq/lambda/SeqTest.java
```diff
@@ -1543,6 +1543,22 @@ public void testIterate() {
     }
 
     @Test
+    public void testIterateWhilePresent() {
+        assertEquals(
+              asList(),
+              Seq.iterateWhilePresent(null, i -> Optional.empty()).toList()
+        );
```
lukaseder: I find this surprising. This should be equal to `asList((Object) null)`. Illustration:

https://github.com/jOOQ/jOOL/commit/0ba503536cefa8c30e0697d2c0c4ce2e8e013f7a

tlinkowski: I understand your surprise - the proposed solution is somewhat inconsistent. The reason I let user provide `null` as `seed` instead of `Optional.empty()` is:
1. `Optional<T>` is essentialy a wrapper for nullable `T`
2. `Optional` is not supposed to be used as a parameter

I see three solutions:
1. call `Objects.requireNonNull` on `seed` (consequences: the resulting `Seq` is always non-empty)
2. change method signature to take `Optional<T> seed` (consequences: we break the convention)
3. leave it as it is (consequences: we're inconsistent in the use of `Optional`s and nullable values)

But what you propose (that `null` `seed` returns a `Seq` with a `null` value) is - in my opinion - the most incoherent solution of all four. If we adopted this, the **only** way `null` could make it into the resulting `Seq` would be as being its first element (because remaining elements are provided by `Optional`, and `Optional` either stores a non-null value or reports it's empty).

lukaseder: Gah, what a mess :) We wouldn't have that trouble with the predicate-based API. Anyway, I'll think about it...

tlinkowski: That's true. Actually, my main objection against `Optional` is that it's not supposed to be used as field type or parameter type. And I don't truly understand why because current approach creates inconsistencies like the one above. If only Java had Kotlin's null-handling...

lukaseder: In this case, it's a return type, not a field/parameter type, so it works along the ideas of the designers, in a way. Even better than Kotlin: Ceylon, with real union types. Your function would return a type like `T|EndOfIteration`

Which reminds me that Java's checked exceptions are really union types (if you squint really hard). Perhaps, we could abort iteration with a checked exception? ;-) (semi-serious)

tlinkowski: Yeah, checked exceptions - I love them ;)

But what I meant by my objection against `Optional` is that I'd like to use it as a parameter (`Optional<T> seed` - solution 2). Actually, when now I look at all these 3 potential solutions, I'm most attracted to this `Optional`-parameter solution. And we already have `Optional` as parameter type, e.g. in `Seq.seq(Optional)`. What do you think?

PS. I don't know about Ceylon and its union types but with Kotlin it would be really nice - we'd have:
```kotlin
fun <T> iterateWhilePresent(seed: T?, generator: (T) -> T?): T
```

lukaseder: An `Optional` is a collection of arity `0..1`, which is why it makes sense to wrap it in a `Seq`. Likewise, JDK 9 will support [`Optional.stream()`](http://download.java.net/java/jdk9/docs/api/java/util/Optional.html#stream--). I wish it even implemented `Iterable`.

In that sense, it might be consistent to allow it as a parameter, but why do that? Why even call `iterate` when we *know* the resulting `Seq` will be empty?I mean, if the `seed` were `Supplier<Optional<T>>`, it might make sense (I wouldn't like it either), but with a hard-wired `Optional`, I think it's better to just not call the method, and value consistency with the existing `iterate()` method higher

tlinkowski: Well, as to why call `iterateWhilePresent` when you *know* that the resulting `Seq` will be empty - it's for the same reason why you call `Seq.seq(Optional)` instead of `optional.isPresent() ? Seq.of(optional.get()) : Seq.empty()`, i.e. for convenience.

But I guess `seed` usually is non-empty so I agree with you that in this particular case consistency with the existing `iteratate()` is more important than robustness.

Solution 1 (`Objects.requireNonNull(seed)`) it is, then, right?

lukaseder: Yeah, I agree with solution 1. It's also the most forward compatible, if we ever want to do something with this `null` seed.

tlinkowski: Done.

lukaseder:
Excellent, thank you very much!

Issue #49, opened 2014-09-22
Add <T1, T2, ..., TN> TupleN<Seq<T1>, Seq<T2>, ..., Seq<TN>> unzipN(Stream<TupleN<T1, T2, ..., TN>>)


lukaseder:
Higher-degree unzipping... Currently, we only support degree 2

PR #305, opened 2017-04-27
Added SeqBuffer<T> class as a replacement for some of Seq.toList() calls

tlinkowski:
I propose a solution to the problems described in #195 for the following methods:
- `crossSelfJoin()` (`crossJoin()` needs to be adapted by @lukaseder in `jOOQ-tools`)
- `inner(Self)Join()`
- `leftOuter(Self)Join()`

The solution for nearly all the remaining methods will consist in applying `Seq.lazy()` to them for which I'll create a separate PR in order to demonstrate its usefulness, as discussed in #302.

---

PS. `SeqBuffer` could be used to create a simpler implementation of `Seq.duplicate`.
PPS. `SeqBuffer` could also be used to implement method `Seq.duplicate(int count)`. Such method would return a `List<Seq<T>>` containing `count` `Seq`s. However, the only application for such method that I can think of is implementing #49 which (to be honest) I don't find very useful either.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -203,11 +203,8 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
      * </pre></code>
      */
     default Seq<Tuple2<T, T>> crossSelfJoin() {
-
-        // This algorithm isn't lazy and has substantial complexity for large argument streams!
-        List<? extends T> list = toList();
-
-        return Seq.crossJoin(seq(list), seq(list));
+        SeqBuffer<T> buffer = SeqBuffer.of(this);
+        return crossJoin(buffer.seq(), buffer.seq());
     }
 
     /**
@@ -244,10 +241,10 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
      */
     default <U> Seq<Tuple2<T, U>> innerJoin(Seq<? extends U> other, BiPredicate<? super T, ? super U> predicate) {
 
-        // This algorithm isn't lazy and has substantial complexity for large argument streams!
-        List<? extends U> list = other.toList();
+        // This algorithm has substantial complexity for large argument streams!
+        SeqBuffer<? extends U> buffer = SeqBuffer.of(other);
 
-        return flatMap(t -> seq(list)
+        return flatMap(t -> buffer.seq()
                            .filter(u -> predicate.test(t, u))
                            .map(u -> Tuple.<T, U>tuple(t, u)))
               .onClose(other::close);
@@ -262,11 +259,8 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
      * </pre></code>
      */
     default Seq<Tuple2<T, T>> innerSelfJoin(BiPredicate<? super T, ? super T> predicate) {
-
-        // This algorithm isn't lazy and has substantial complexity for large argument streams!
-        List<? extends T> list = toList();
-
-        return Seq.<T>seq(list).innerJoin(seq(list), predicate);
+        SeqBuffer<T> buffer = SeqBuffer.of(this);
+        return buffer.seq().innerJoin(buffer.seq(), predicate);
     }
 
     /**
@@ -303,10 +297,10 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
      */
     default <U> Seq<Tuple2<T, U>> leftOuterJoin(Seq<? extends U> other, BiPredicate<? super T, ? super U> predicate) {
 
-        // This algorithm isn't lazy and has substantial complexity for large argument streams!
-        List<? extends U> list = other.toList();
+        // This algorithm has substantial complexity for large argument streams!
+        SeqBuffer<? extends U> buffer = SeqBuffer.of(other);
 
-        return flatMap(t -> seq(list)
+        return flatMap(t -> buffer.seq()
                            .filter(u -> predicate.test(t, u))
                            .onEmpty(null)
                            .map(u -> Tuple.<T, U>tuple(t, u)))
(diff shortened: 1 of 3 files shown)


tlinkowski:
I added a new implementation of `Seq.splitAt` [#308] that uses the proposed `SeqBuffer`.

lukaseder:
That looks very interesting, thank you very much for your suggestion. I'll review this later (hopefully in the next 2 days). We'll definitely need a reusable `SeqBuffer`.

src/main/java/org/jooq/lambda/SeqBuffer.java
```diff
@@ -0,0 +1,110 @@
...
+
+    static <T> SeqBuffer<T> of(Spliterator<T> spliterator) {
+        if (spliterator instanceof SeqBuffer.BufferSpliterator) {
+            return ((SeqBuffer<T>.BufferSpliterator) spliterator).parentSeqBuffer(); // reuse existing SeqBuffer
+        }
+        return new SeqBuffer<>(spliterator);
+    }
+
+    private final Spliterator<T> source;
+    private final List<T> buffer = new ArrayList<>();
```
lukaseder: Hmm, this buffer is not thread safe. It may well be that one consumer of a buffered result consumes the buffer while another consumer produces new elements at the same time. E.g., this could happen in your `splitAt()` implementation suggestion here: https://github.com/jOOQ/jOOL/pull/305/commits/b345bbb8244e727d01b7a59ee10835723e5da72c

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -9688,15 +9688,8 @@ public T next() {
...
-            .zipWithIndex()
-            .partition(t -> t.v2 < position)
-            // Explicit type parameters to work around this Eclipse compiler bug:
-            // https://bugs.eclipse.org/bugs/show_bug.cgi?id=455945
-            .map((v1, v2) -> Tuple.<Seq<T>, Seq<T>>tuple(
-                v1.map(t -> t.v1),
-                v2.map(t -> t.v1)
-            ));
+        SeqBuffer<T> buffer = SeqBuffer.of(stream);
+        return tuple(buffer.seq().limit(position), buffer.seq().skip(position));
```
lukaseder: Much nicer, indeed. Only caveat: Thread safety in the current `SeqBuffer` implementation

tlinkowski: As a side note: do you realize that the current implementation of `Seq.splitAt` (as well as `Seq.partition`) is not thread safe? :)

src/main/java/org/jooq/lambda/SeqBuffer.java
```diff
@@ -57,6 +63,17 @@ private SeqBuffer(Spliterator<T> source) {
...
 
+    private static <T> List<T> initBuffer(Spliterator<T> source) {
+        long estimateSize = source.estimateSize();
+        if (estimateSize == Long.MAX_VALUE) // infinite/unknown size
+            return new ArrayList<>();
+
+        if (estimateSize > MAX_ARRAY_SIZE)
+            throw new IllegalArgumentException("Stream is too long to be buffered: " + estimateSize);
+
+        return new ArrayList<>((int) estimateSize);
```
lukaseder: Hm, no I don't agree with this. If we create a large capacitied array list, we're back to the original solution that you wanted to avoid: Up-front large memory consumption for intermediate buffers. Imagine:

```java
Seq.seq(Collections.nCopies(1000000, "value")).splitAt(999999).v1.limit(1).forEach(System.out::println);
```

The `splitAt()` operation would create a large array list only to consume its first element...

tlinkowski: Ha, you're right! I'll revert it.

PS. We wouldn't be entirely back to square one because I mainly wanted to avoid up-front processor consumption (i.e. consuming entire `Seq`) but you're absolutely right that it's best to also avoid up-front memory consumption.

lukaseder:
Thanks for the fix. Now, I wonder again if it would be possible to squash all commits into one for this PR (GitHub has such a feature). That would make the final review much simpler for me.

tlinkowski:
Done.

PS. I couldn't find such feature (available to the PR author) anywhere on GitHub so I squashed them manually.

lukaseder:
> PS. I couldn't find such feature (available to the PR author) anywhere on GitHub so I squashed them manually.

Oh, interesting. That was a wrong assumption then, sorry. I can indeed squash and merge, but not squash, review again, then merge... Will send a feature request to GitHub.

Thanks very much!

src/main/java/org/jooq/lambda/Seq.java
billoneil: I was under the impression that `Seq` wasn't thread safe to begin with or was it just that it will not be a parallel stream?

lukaseder: There's currently no guarantee in the API, as there are too many flaws, still. But in principle, the results from methods like `splitAt()` (which produce several Seq) should be thread safe. Or at least, we should have an option for them to be thread safe.

Perhaps this is a concept that should be reviewed more globally.

Note that thread safety and parallelism aren't the same thing in this context. Stream's parallelism allows for parallel processing of operations like `map()` or `filter()`. By keeping operations independent of one another, parallelism can drastically speed up the processing of a stream. Thread safety in our context just means that consuming two things (e.g. the `splitAt()` results) on different threads, we don't want to get wrong results.

But again, perhaps we should implement this more thoroughly, with a specific thread safety flag...

lukaseder:
Thanks again for this change. I'll review this later in detail, and will let you know.

src/main/java/org/jooq/lambda/Seq.java
tlinkowski: Well, `Seq.splitAt` was directly based on `Seq.partition` :) And `Seq.partition` definitely isn't thread-safe because both `Seq`s write to `buffer1` and `buffer2` without any synchronization.

tlinkowski: @billoneil If I understand it right I think @lukaseder distinguishes between two things:
- single `Seq` is sequential so as a consequence it's not thread safe and two threads cannot operate on the same `Seq`
- there seems to be no restriction to consuming one `Seq` on one thread and another `Seq` on the other thread even if they both "derive" from the same `Seq` (hence need for thread safety in such cases)

billoneil: Ah I wasn't thinking, makes sense.

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -9165,43 +9156,8 @@ public R next() {
...
-                        ahead[0] = null;
-
-                    return value;
-                }
-            }
-        }
-
-        return tuple(seq(new Duplicate()), seq(new Duplicate()));
+        SeqBuffer<T> buffer = SeqBuffer.of(stream);
+        return tuple(buffer.seq(), buffer.seq());
```
tlinkowski: You're absolutely right. On the other hand, current implementation is thread safe thanks to `SeqBuffer`, and the previous one wasn't.

All in all, I'd be more inclined to this simple implementation but I guess I'd leave an appropriate comment in this method in case someone reports any memory-related issues in the future.

PR #307, opened 2017-04-28
Make Seq.flatMap() lazy

tlinkowski:
@lukaseder, I know I've been flooding you with PRs recently, sorry... But this one is a result of my astonishment that `Stream.flatMap` is not a fully lazy operation!

Some (me included) consider this behavior to be a bug: https://bugs.openjdk.java.net/browse/JDK-8075939 But maybe there was some rationale for `Stream.flatMap` not to be fully lazy because `Stream`s may be parallel. However, on `sequential` streams like `Seq`, it seems there is absolutely no justification for `flatMap` to be so eager (but - to quote you - "I'm happy to be proven wrong").

So this PR proposes a `Seq.flatMapLazily()` method which corresponds directly to `Seq.flatMap()` but is fully lazy. I spent an entire day writing this but it was fun :)

Theoretically, we could venture as far as modifying `Seq.flatMap` to call this lazy implementation. On the one hand, it would be better than introducing a new method in `Seq` that doesn't (functionally) do anything new. On the other hand, it seems quite risky because it could potentially break a loooot of client code. Although, for testing on some large codebase (like jOOQ), it'd be a nice thing to do (I mean, to modify `flatMap`, and then run all the tests on the codebase, and see where they fail).

The tests that I've provided in this PR are rather scarcy, and I feel this feature should be **tested much more extensively** before being accepted. I didn't write more because I didn't want to spend too much time on it up front.

I just hope you'll seriously consider and then (hopefully) revise this request (no hurry, of course) because:
- current implementation of sequential `flatMap` looks like a potentially huge waste of resources
- I'd make use of it in my project, and to use it conveniently it needs to be an interface method :)

### DIFF
--- a/src/main/java/org/jooq/lambda/LazyFlatMapper.java
+++ b/src/main/java/org/jooq/lambda/LazyFlatMapper.java
@@ -19,6 +19,9 @@ import java.util.stream.Stream;
  */
 final class LazyFlatMapper<T, R> {
 
+    /**
+     * Flat maps this stream using given <code>mapper</code> but in a fully lazy way.
+     */
     static <T, R> Seq<R> flatMapLazily(Stream<? extends T> stream, Function<? super T, ? extends Stream<? extends R>> mapper) {
         return new LazyFlatMapper<T, R>(mapper).flatMapLazily(stream);
     }
--- a/src/main/java/org/jooq/lambda/LazyFlatMapper.java
+++ b/src/main/java/org/jooq/lambda/LazyFlatMapper.java
@@ -1,0 +1,121 @@
+package org.jooq.lambda;
+
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * Special class that implements a fully lazy <code>flatMap</code> operation.
+ * Default <code>flatMap</code> in <code>Stream</code> is only a partially lazy operation:
+ * @link http://stackoverflow.com/questions/29229373/why-filter-after-flatmap-is-not-completely-lazy-in-java-streams
+ *
+ * This class should not be instantiated outside of the provided static <code>flatMapLazily</code> method.
+ *
+ * @see Stream#flatMap
+ *
+ * @author Tomasz Linkowski
+ */
+final class LazyFlatMapper<T, R> {
+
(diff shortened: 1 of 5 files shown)


lukaseder:
> @lukaseder, I know I've been flooding you with PRs recently, sorry...

Don't be, it's great to have "fresh wind" in the project :) I might just take some time replying, as jOOQ is keeping me busy. But we're not in a hurry.

> (but - to quote you - "I'm happy to be proven wrong").

No need, here. `Seq.flatMap()` is eager only because it was not a priority to think about how to make it lazy (yet). Much like linear-time window functions aren't implemented yet, they're mostly quadradic.

But let's focus on `flatMap()` first. I'm sure it's much easier to optimise.

> Theoretically, we could venture as far as modifying Seq.flatMap to call this lazy implementation.

Yes. I definitely don't want to introduce this "lazy" concept. Everything should be lazy in `Seq`. The only reason why something isn't lazy is because it isn't lazy yet.

> On the other hand, it seems quite risky because it could potentially break a loooot of client code. 

Yes. Welcome to library development :)

But what are you afraid of? Anything in particular?

> Although, for testing on some large codebase (like jOOQ), it'd be a nice thing to do (I mean, to modify flatMap, and then run all the tests on the codebase, and see where they fail).

jOOQ doesn't make much use of jOOλ. There are a few integration tests that use it, but they don't make excessive use of `flatMap()`.

There's definitely room for tons of new `flatMap()` unit tests in jOOλ, though :)

So, yes. I agree: Let's go ahead and look at this.

tlinkowski:
>> Theoretically, we could venture as far as modifying Seq.flatMap to call this lazy implementation.
>
> Yes. I definitely don't want to introduce this "lazy" concept. Everything should be lazy in Seq. The only reason why something isn't lazy is because it isn't lazy yet.

I'm glad you prefer this riskier but cleaner solution :)

> But what are you afraid of? Anything in particular?

Well, not really - just that some important implementation detail might have escaped me, or something like that. Fortunately, synchronization does not play a role here (as far as I understand) so it feels much easier to implement it right.

> There's definitely room for tons of new flatMap() unit tests in jOOλ, though :)

Yeah, that's true. I'll write some more for sure but I wonder who'll write the rest till we have "tons" ;)

lukaseder:
> Yeah, that's true. I'll write some more for sure but I wonder who'll write the rest till we have "tons" ;)

Well, we're absolutely not in a rush to implement this...

lukaseder:

Issue #309, opened 2017-04-30
Seq.limitUntil sometimes fetches too many elements from source


tlinkowski:
When calling `Seq.of(1, 2, 3, 4, 5).limitUntil(i -> i > 1).iterator()`, and then calling `Iterator.next()` until `!Iterator.hasNext()`, two elements are fetched from the source `Seq` (`1` that fails the `Predicate`, and `2` that satisfies the `Predicate`), which is correct behavior.

However, when calling `Iterator.hasNext()` one more time (which is redundant but valid), one more element (`3`) is fetched from the source `Seq`, which is incorrect behavior.

This behavior breaks some code, e.g. the following:
```java
interface Node {
  @Nullable Node parent();
}

static Seq<Node> ancestorSeq(Node node) {
  return Seq.iterate(node, Node::parent).limitUntil(Objects::isNull);
}
```

PS. All the other methods (`limitUntilClosed`, `skipUntil`, `skipUntilClosed`) work correctly.

tlinkowski:
Oh, I found out where this defect comes from: in `Seq.limitUntil` the condition `!test[0]` must be tested before (and not after) calling `delegate.tryAdvance`:
```java
!test[0] && delegate.tryAdvance(t -> {
    if (!(test[0] = predicate.test(t)))
        action.accept(t);
})
```

lukaseder:
Thank you very much for your analysis. I can confirm that your suggested solution works. Interestingly, `limitUntilClosed` does it correctly with the proper order of booleans in the expression...

Issue #308, opened 2017-04-30
Seq.splitAt implementation is inefficient


tlinkowski:
Iterating `Seq.of(1, 2, 3, 4, 5, 6).splitAt(2).v1()` requires consumption of entire `Seq`, and not only consumption of the first 2 elements (i.e. until given `position` is reached).

I propose the following implementation using `SeqBuffer` proposed in #305:
```java
    static <T> Tuple2<Seq<T>, Seq<T>> splitAt(Stream<? extends T> stream, long position) {
        SeqBuffer<T> buffer = SeqBuffer.of(stream);
        return tuple(buffer.seq().limit(position), buffer.seq().skip(position));
    }
```

lukaseder:
OK this depends on `SeqBuffer`. I'll look at that first: https://github.com/jOOQ/jOOL/issues/305

lukaseder:
Hmm, I think you haven't sent a PR for these changes yet. Would you mind doing so, so you can get credit?

tlinkowski:
Sure!

tlinkowski:
Hm, I just realized that I implemented it inside [#305] (in the squashed commit 61e75370). Unfortunately, I forgot to add "[#308] more efficient Seq.splitAt implementation" to the commit message.

lukaseder:
Oh yes indeed, thanks very much. The test also runs. I've activated it again.

PR #310, opened 2017-04-30
Added SeqTest.verifyIteratorValidity() method

tlinkowski:
I propose a new private method in `SeqTest`:
```java
void verifyIteratorValidity(int expectedCount, Seq<T> source, Function<Seq<T>, Seq<?>> operation)
```

This method verifies whether our non-terminal `Seq` operations (e.g. `limitWhile`, `partition`, `splitAt`) behave as expected, i.e.:
- whether every call to `Spliterator.tryAdvance` that returns `true` actually calls given `action`
- whether the number of elements fetched from the `source` sequence is equal to `expectedCount`
- whether calling redundantly `Iterator.hasNext` does not fetch any extra element from the source sequence (such calls are valid and should be "safe")

Using this method I detected two defects: #308 and #309.


lukaseder:
A late thank you very much for this suggestion. I agree with these tests, very useful.

lukaseder:
Again, I really like this utility. It will help us validate the laziness of a lot of future features... Thanks again!

PR #312, opened 2017-05-03
Include methods to retrieve values inside the interface Tuple

andrebreves:
Includes `v1()`, `v2()`, ... inside interface `Tuple` for each tuple concrete type, so you can reference methods using the interface (`Tuple::v1`, `Tuple::v2`, ...), making constructs like:
```java
.stream()
.map(x -> Tuple.tuple(x.a(), x.b()))
.sorted(comparing(Tuple2::v2))
.map(Tuple2::v1)
``` 
Cleaner and more refactor friendly:
```java
.stream()
.map(x -> Tuple.tuple(x.a(), x.b()))
.sorted(comparing(Tuple::v2))
.map(Tuple::v1)
```


billoneil:
Personally I think this could introduce more bugs and I would prefer to explicitly change the types if my model changes.

What if your code changes from

```.map(x -> Tuple.tuple(x.a(), x.b()))``` to ```.map(x -> Tuple.tuple(x.a(), x.y(), x.b()))```

In your example it would change the functionality to now be sorting by x.y instead of x.b. Probably not what you intended. The existing code would give you a compile time error and make you update it to `Tuple3`

lukaseder:
Thank you very much for the suggestion. I perfectly understand the pain point here and don't fully disagree, but I'm a bit wary of compile time speed given that compiling jOOλ still crashes Eclipse. I wonder if the added value really justifies the risk of adding new compiler challenges here...

lukaseder:
(btw: thanks for the PR work. We have a code generator for all these `T1` .. `T16` ;) (but not open source))

nnordrum:
@lukaseder what's the preferred method to get changes to your generated output? Also, do you have one for Function (before I go and manually type out way too many additions for a PR).

nnordrum:
@lukaseder could that maybe be documented in a more prominent spot than deep in the comments of #312? :smile:

lukaseder:
@nnordrum : Yeah, good idea ;) I'll create a CONTRIBUTING.md file. Let's hope people will see the link

lukaseder:

Reviewing this feature request, I don't think this is a feature worth adding

Issue #313, opened 2017-05-03
Add Seq.skipLast(int n) and Seq.limitLast(int maxSize)


tlinkowski:
I propose `Seq.skipEnd(int n)` and `Seq.limitEnd(int maxSize)` because I just came across the need to skip the last element from the `Seq` and couldn't do that easily (i.e. without resorting to `Seq.reverse().skip(1).reverse()` which is extremely inefficient).

They would correspond to `skip(long n)`  and `limit(long maxSize)` but they would be `int-`based instead of `long`-based. I mentioned these methods in #241 but didn't create a separate issue then:
> These methods could be implemented with a circular buffer of size `maxSize/n` (e.g. an `ArrayDeque` could be used for this purpose). I guess it'd get harder if we wanted to have `long`s instead of `int`s as the arguments.

PS. Alternative API approach: allow negative values in current `Seq.skip` and `Seq.limit` methods - a negative value would mean "count from the right".

lukaseder:
Thanks for your suggestion. Hmm, I'm not so sure if I can see the use-case here. Would you mind describing it? Also, do you see an implementation that would be *much* better (i.e. better complexity) than the one you already proposed?

tlinkowski:
Well, general use-case is analogous to normal `limit` and `skip` :) Of course, perhaps the most common use-case  of normal `limit` and `skip` is "pagination" (like in SQL). But apart from that, they have normal usages like "I don't want first 5 elements" or "I want only first 5 elements". So why shouldn't you be able to say "I don't want last 5 elements" or "I want only last 5 elements"? BTW. maybe the name should be `limitLast` and `skipLast` then.

As to my specific use-case - I got a `Seq`, and needed to verify all elements except the last. Finally, however, the design changed and it turned out I needed to verify all elements except the first which is straightforward.

As for the implementation - as far as I understand **no** better implementation is possible (not even slightly). In order to make a decision (whether to begin outputting - in case of `limitEnd`, or whether to stop outputting - in case of `skipEnd`) you **must** know `maxSize/n` elements beforehand, and to know them means to buffer them.

Actually, I'm a bit surprised about your complexity concerns. As for the memory requirements, an operation like `partition(n)` or `splitAt(n)` requires caching entire `Seq` no matter the `n`. And `skipEnd(n)`/`limitEnd(n)` require caching up to `n` elements only. But maybe you're concerned about something else than memory - if so, please, specify because I cannot see any other problems here.

[...]

PR #314, opened 2017-05-10
Added Seq.skipLast(long n) and Seq.limitLast(long maxSize)

tlinkowski:
See #313.

### DIFF
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -3233,6 +3233,21 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
     }
 
     /**
+     * Returns a stream without its <code>n</code> last elements,
+     * or an empty stream if it has less than <code>n</code> elements.
+     * <p>
+     * <code><pre>
+     * // (1, 2, 3)
+     * Seq.of(1, 2, 3, 4, 5).skipLast(2)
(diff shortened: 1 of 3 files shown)


src/main/java/org/jooq/lambda/SequentialSpliterator.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Special subinterface of <code>Spliterator</code> that provides default implementations
+ * for all the methods that are not relevant to sequential traversal.
+ *
+ * @see java.util.stream.BaseStream#sequential
+ *
+ * @author Tomasz Linkowski
+ */
+@FunctionalInterface
+public interface SequentialSpliterator<T> extends Spliterator<T> {
```
lukaseder: Hmm, interesting - good idea. Perhaps we should put that in a separate issue, though. I'll decide later, when it's clear if this type really needs to be public

tlinkowski: For now, I just placed it inside `TailBuffer`. But maybe it doesn't need to be public, I must say I just added it as public without much consideration.

lukaseder: I see. As always, I prefer this not to be public, then :)

lukaseder:

lukaseder:
I'm not sure we need this in the API

PR #315, opened 2017-05-17
Added Seq.peekThrowable(Consumer<Throwable>)

tlinkowski:
See #91 and #299.

@lukaseder, you should be aware that the implementation I provided here (which is the same as proposed in #299) does not have the exact semantics you expected. Namely, the following test passes:
```java
// assert throws PeekedX EVEN when "peekThrowable" BEFORE "throw" but on the SAME Spliterator
assertThrows(PeekedX.class, () -> Seq.of(1, 2, 3)
    .peekThrowable(throwable -> {
        throw new PeekedX();
    })
    .peek(i -> {
        if (i == 2)
            throw new X();
    })
    .toList());
```

which means that `peekThrowable` works not only upstream but also downstream. The reason for this is (as far as I understand) that the `action` passed to the top-most (source) `Spliterator.tryAdvance(Consumer<T> action)` represents (in a way) all the cumulated operations (up to the terminal operation).

However, calling `peekThrowable` more up or down the stream can make a difference because we can catch two types of `Throwable`s:
1. Thrown inside `action.apply(T)` - these are caught both upstream and downstream
2. Thrown in the remaining parts of `Spliterator.tryAdvance()` - these are caught only upstream

### DIFF
--- a/src/test/java/org/jooq/lambda/SeqTest.java
+++ b/src/test/java/org/jooq/lambda/SeqTest.java
@@ -987,6 +987,50 @@ public class SeqTest {
         assertEquals(asList(2, 3), Seq.of(2, 3).onEmptyThrow(() -> new X()).toList());
     }
 
+    @Test
+    public void testPeekThrowable() {
+        class PeekedX extends RuntimeException {}
+
+        Function<Integer, Consumer<Integer>> throwXUpon = i -> j -> {
+            if (i.equals(j)) {
+                throw new X();
+            }
+        };
+        Consumer<Throwable> rethrowAsPeekedX = ex -> {
+            throw new PeekedX();
+        };
+
+        // assert throws X
+        assertThrows(X.class, () -> Seq.of(1, 2, 3)
+              .peek(throwXUpon.apply(2))
+              .toList());
+
+        // assert throws PeekedX when "peekThrowable" AFTER "throw"
+        assertThrows(PeekedX.class, () -> Seq.of(1, 2, 3)
+              .peek(throwXUpon.apply(2))
+              .peekThrowable(rethrowAsPeekedX)
+              .toList());
+
+        // assert throws PeekedX EVEN when "peekThrowable" BEFORE "throw" but on the SAME Spliterator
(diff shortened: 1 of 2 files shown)


src/test/java/org/jooq/lambda/SeqTest.java
```diff
@@ -988,7 +988,7 @@ public void testOnEmpty() throws X {
     }
 
     @SuppressWarnings("serial")
-    class X extends Exception {}
+    static class X extends RuntimeException {}
```
lukaseder: Hmm, there was an implicit test to validate that things work for checked exceptions, too. I'll revert this part, rename `X` to `Checked` and add a new `Unchecked` exception for your test after mergin.

lukaseder:
Thanks a lot for your PR!

> does not have the exact semantics you expected

Maybe I didn't write it down this way, but the example is exactly what I would expect.

> Thrown inside action.apply(T) - these are caught both upstream and downstream

Could you give an example for this?

tlinkowski:
Well, in #299 you wrote:
```java
Seq.seq(source)
   .filter(op1)
   .filter(op2)
   .peekThrowable(e -> ... /* only exceptions from op1 and op2 */)
   .filter(op3)
   .peekThrowable(e -> ... /* exceptions from op1, op2, or op3 */)
   .collect(collector);
```

but we really have:
```java
Seq.seq(source)
   .filter(op1)
   .filter(op2)
   .peekThrowable(e -> ... /* all exceptions from op1 and op2 + most exceptions from op3 */)
   .filter(op3)
   .peekThrowable(e -> ... /* all exceptions from op1, op2, or op3 */)
   .collect(collector);
```

for example:
```java
Seq.of(1, 2, 3)
   .filter(i -> i > 1)
   .filter(i -> i <= 3)
   .peekThrowable(System.out::println) // will print RuntimeException
   .filter(i -> {
     if (i == 3)
         throw new RuntimeException();
     return i < 3;
  })
  .toList();
```

lukaseder:
Oh, interesting, thanks for clarifying. Then I don't agree with the status quo. Will revert. That's very surprising and doesn't work like `peek()`...

Issue #296, opened 2017-03-11
chunked(long), chunked(Predicate<T>)


billoneil:
I'm looking for a way to take a `Seq<T>` and transform it to a `Seq<List<T>>` for batch processing while keeping lazy evaluation. (I am streaming and processing large files from S3 then pushing them back to S3)

I found the following and have two concerns. http://stackoverflow.com/a/30662609
1. `Note that this solution unnecessarily stores the whole input stream to the intermediate Map (unlike, for example, Ben Manes solution)` Is this accurate?
2. I believe the groupBy would exhaust the entire stream. In my case the file may not fit in memory so this would not work if that is the case.

Would the above be worth creating a helper for? Or would it be worth making something similar to Guavas `Iterables.partition`

billoneil:
Example 
```
Seq.of(1,2,3,4,5,6).grouped(3);
// ((1,2,3), (4,5,6))
```

I tried looking into the Window functions but I'm not sure if that would work.

lukaseder:
Thanks for your suggestion

1) Yes that's accurate. The solutions I suggested on Stack Overflow are workarounds that work well for small data sets, but don't scale as they're not lazy
2) GroupBy is a terminal operation, indeed, so the stream is consumed
3) Grouped already exists but it might not be a good candidate here, as it also risks consuming the entire stream
4) Window functions currently aren't lazy (enough)

So, yes, perhaps a helper might be worth exploring. Now about naming :) It looks like it's the opposite of `flatMap()` if you neglect type formalisms. `bufferMap()` perhaps? I'm not so fan of recycling the term `group` as the existing `group` notion groups by value, not index.

Also, the new functionality should probably ship with two flavours. "grouping" by index, and by predicate (e.g. a predicate returning true at the last element of a "group")

ben-manes:
How about `splitBy(int)` and `splitBy(pred)`? This might feel like a plural version of `splitAt` which does a singular split.

lukaseder:
Nice thinking, although the return types are fundamentally different, so reusing the term might seem surprising. How do Haskell et al. call this operation?

ben-manes:
Haskell calls it [chunksOf](https://hackage.haskell.org/package/split-0.2.3.1/docs/Data-List-Split.html#v:chunksOf).

lukaseder:
[...]

PR #320, opened 2017-07-25
[#296] Added chunked(long), chunked(Predicate<T>)

billoneil:
#296
I have two of the implementations working and they should be truly lazy.

- [x] chunked(long chunkSize)
- [x] chunked(Predicate<T> predicate)
- [ ] chunked(Predicate<T> predicate, boolean excludeDelimiter) ?
- [ ] chunkedWithIndex(BiPredicate<T, Long> biPredicate)

I'm not sure happy with how the code turned out. The two iterators are a little dependent on each other. Maybe a better approach would be to make a more top level peeking iterator instead of hacking it like I did.

### DIFF
--- a/src/main/java/org/jooq/lambda/Iterators.java
+++ b/src/main/java/org/jooq/lambda/Iterators.java
@@ -1,0 +1,116 @@
+package org.jooq.lambda;
+
+import java.util.Iterator;
+import java.util.function.Predicate;
+
+/**
+ * Created by billoneil on 7/26/17.
+ */
+public class Iterators {
+    private enum PredicateCheckState { NOT_TESTED, MORE, ONE_MORE, NO_MORE }
+
+    static <E> PeekingIterator<E> peeking(Iterator<E> iterator) {
+        return new PeekingIteratorImpl<E>(iterator);
+    }
+
+    static <E> Iterator<E> takeWhileIterator(Iterator<E> iterator, Predicate<E> predicate) {
+        return takeWhileIterator(peeking(iterator), predicate);
+    }
+
+    static <E> Iterator<E> takeWhileIterator(PeekingIterator<E> iterator, Predicate<E> predicate) {
+        return new Iterator<E>() {
+
+            private final PeekingIterator<E> peeking = iterator;
+            private PredicateCheckState state = PredicateCheckState.NOT_TESTED;
+
+            @Override
+            public boolean hasNext() {
+                // Short circuit since its not possible
+                if (!peeking.hasNext()) {
+                    state = PredicateCheckState.NOT_TESTED;
+                    return false;
+                }
+
+                // We want the predicate test to be deterministic per element
+                // so we need to cache the result until we advance the iterator.
+                // An example predicate would be a counting predicate.
+                if (PredicateCheckState.NOT_TESTED == state) {
+                    boolean stop = predicate.test(peeking.peek());
+                    if (stop) {
+                        state = PredicateCheckState.ONE_MORE;
+                    } else {
+                        state = PredicateCheckState.MORE;
+                    }
+                }
+
+                if (state == PredicateCheckState.NO_MORE) {
+                    return false;
+                }
+                return true;
+            }
+
+            @Override
+            public E next() {
+                if (PredicateCheckState.ONE_MORE == state) {
+                    state = PredicateCheckState.NO_MORE;
+                } else {
+                    state = PredicateCheckState.NOT_TESTED;
+                }
+                return peeking.next();
+            }
+        };
+    }
+
+    static class PeekingIteratorImpl<E> implements PeekingIterator<E> {
+        private final Iterator<E> delegate;
+        private E headElement;
+        private boolean hasPeeked;
+
+        PeekingIteratorImpl(Iterator<E> delegate) {
+            this.delegate = delegate;
+            this.headElement = null;
+            this.hasPeeked = false;
+        }
+
+        @Override
+        public E peek() {
+            if (hasPeeked) {
+                return headElement;
+            }
+            headElement = delegate.next();
+            hasPeeked = true;
+            return headElement;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return hasPeeked || delegate.hasNext();
+        }
+
+        @Override
+        public E next() {
+            if (hasPeeked) {
+                hasPeeked = false;
+                return headElement;
+            }
+            return delegate.next();
+        }
+    }
+
+    static <T> Predicate<T> countingPredicate(long count) {
+        if (count < 1) {
+            throw new IllegalArgumentException("count must be greater than 1");
+        }
+        long[] countArray = new long[1];
+        countArray[0] = 0L;
+        long iterations = count -1;
+        return (T t) -> {
+            if (iterations == countArray[0]) {
+                countArray[0] = 0L;
+                return true;
+            }
+            countArray[0] = countArray[0] + 1;
+            return false;
+        };
+    }
+}
--- a/src/main/java/org/jooq/lambda/PeekingIterator.java
+++ b/src/main/java/org/jooq/lambda/PeekingIterator.java
@@ -1,0 +1,13 @@
+package org.jooq.lambda;
+
+import java.util.Iterator;
+
+public interface PeekingIterator<E> extends Iterator<E> {
+    /**
+     * Returns the current head of the iterator without consuming it.
+     * Multiple sequential calls to {@code peek()} will return the same element.
+     * @return The current head of the iterator
+     * @throws java.util.NoSuchElementException if there are no more elements
+     */
+    E peek();
+}
--- a/src/main/java/org/jooq/lambda/Seq.java
+++ b/src/main/java/org/jooq/lambda/Seq.java
@@ -3397,6 +3397,30 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
     }
 
     /**
+     * Chunk a Stream based on a predicate
+     * <p>
+     * <code><pre>
+     * // ((1,1,2), (2), (3,4))
+     * Seq.of(1, 1, 2, 2, 3, 4).chunked(n -> n %2 == 0)
+     * </pre></code>
+     */
+    default Seq<Seq<T>> chunked(Predicate<T> predicate) {
+        return chunked(this, predicate);
+    }
+
+    /**
+     * Chunk a Stream based on a number of elements
+     * <p>
+     * <code><pre>
+     * // ((0,1,2), (3,4,5), (6,7,8), (9))
+     * Seq.of(0, 1, 2, 3, 4, 5, 6, 7, 8, 9).chunked(3)
+     * </pre></code>
+     */
+    default Seq<Seq<T>> chunked(long chunkSize) {
+        return chunked(this, Iterators.countingPredicate(chunkSize));
+    }
+
+    /**
      * Returns a limited interval from a given Stream.
      * <p>
      * <code><pre>
@@ -9710,6 +9734,36 @@ public interface Seq<T> extends Stream<T>, Iterable<T>, Collectable<T> {
         return tuple(it.hasNext() ? 
                 Optional.of(it.next()) : 
                 Optional.empty(), seq(it));
+    }
+
+    /**
+     * Chunk a Stream based on a predicate
+     * <p>
+     * <code><pre>
+     * // ((1,1,2), (2), (3,4))
+     * Seq.of(1, 1, 2, 2, 3, 4).chunked(n -> n %2 == 0)
+     * </pre></code>
+     */
+    static <T> Seq<Seq<T>> chunked(Stream<T> stream, Predicate<T> predicate) {
+        // This iterator can get in a bad state if you call next
+        // multiple times without consuming the intermediate Seq.
+        // It should be safe since its not exposed. A fix could
+        // be to force has next to make sure the current Seq was iterated.
+        Iterator<Seq<T>> it = new Iterator<Seq<T>>() {
+            private final PeekingIterator<T> peeking = Iterators.peeking(stream.iterator());
+
+            @Override
+            public boolean hasNext() {
+                return peeking.hasNext();
+            }
+
+            @Override
+            public Seq<T> next() {
+                return Seq.seq(Iterators.takeWhileIterator(peeking, predicate));
+            }
+        };
+
+        return Seq.seq(it);
     }
 
     // Methods taken from LINQ
--- a/src/test/java/org/jooq/lambda/IteratorsTest.java
+++ b/src/test/java/org/jooq/lambda/IteratorsTest.java
@@ -1,0 +1,91 @@
+package org.jooq.lambda;
+
+import org.junit.Test;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.function.Predicate;
+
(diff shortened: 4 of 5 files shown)


billoneil:
Refactored it a bit to clean it up. The peeking iterator seems to work. Guava's version special cases remove not sure if that's necessary if its not publicly exposed.

lukaseder:
Thank you very much for your suggestion. I'll review shortly

src/main/java/org/jooq/lambda/Iterators.java
```diff
@@ -0,0 +1,116 @@
+package org.jooq.lambda;
+
+import java.util.Iterator;
+import java.util.function.Predicate;
+
+/**
+ * Created by billoneil on 7/26/17.
+ */
+public class Iterators {
```
lukaseder: Can you please make this class package-private

src/main/java/org/jooq/lambda/PeekingIterator.java
```diff
@@ -0,0 +1,13 @@
+package org.jooq.lambda;
+
+import java.util.Iterator;
+
+public interface PeekingIterator<E> extends Iterator<E> {
```
lukaseder: Same here. This is an internal type and shouldn't be public, in my opinion.

lukaseder: In fact, I wonder if this needs to be an interface, given we only have a single implementation so far...

src/main/java/org/jooq/lambda/Seq.java
```diff
@@ -3397,6 +3397,30 @@ public R next() {
     }
 
     /**
+     * Chunk a Stream based on a predicate
+     * <p>
+     * <code><pre>
+     * // ((1,1,2), (2), (3,4))
+     * Seq.of(1, 1, 2, 2, 3, 4).chunked(n -> n %2 == 0)
```
lukaseder: Hmm, I think we should exclude the chunk boundaries by default if using a predicate, and include them only explicitly. When including, the question is whether the boundary should belong to:

1. The ending chunk `((1, 1, 2), (2), (3, 4))`
2. The beginning chunk `((1, 1), (2), (2, 3), (4))`
3. Both chunks (?)
4. No chunk

Given that there is no clear preference between 1/2 and maybe even 3, the default must be 4. So, an enum might be a reasonable additional argument for this, where "no chunk" would be the default.

src/test/java/org/jooq/lambda/IteratorsTest.java
```diff
@@ -0,0 +1,91 @@
...
+        assertEquals(1, (int) peeking3.peek());
+        assertEquals(1, (int) peeking3.next());
+        assertEquals(2, (int) peeking3.next());
+        assertEquals(3, (int) peeking3.peek());
+        assertEquals(3, (int) peeking3.peek());
+        assertEquals(3, (int) peeking3.next());
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testCountingPredicateThrowsIllegalArg() {
```
lukaseder: I see, this is probably the reason why you made the class public...

1. If the tests are placed in the same package, you can access package-private classes.
2. I'm not too fan of testing internals with unit tests. The internals will be tested transitively by testing the chunked API. With tests on internals, refactoring those internals will be much harder...

billoneil:
I realized some naming is off `takeWhileIterator` should probably be `takeUntil`.

src/main/java/org/jooq/lambda/Iterators.java
```diff
@@ -0,0 +1,116 @@
...
+
+    static <E> PeekingIterator<E> peeking(Iterator<E> iterator) {
+        return new PeekingIteratorImpl<E>(iterator);
+    }
+
+    static <E> Iterator<E> takeWhileIterator(Iterator<E> iterator, Predicate<E> predicate) {
+        return takeWhileIterator(peeking(iterator), predicate);
+    }
+
+    static <E> Iterator<E> takeWhileIterator(PeekingIterator<E> iterator, Predicate<E> predicate) {
```
lukaseder: I have a feeling that we keep writing this iterator :) Perhaps, can this be refactored with existing code?

src/main/java/org/jooq/lambda/Iterators.java
billoneil: I meant to make these package-private oops.

src/main/java/org/jooq/lambda/Iterators.java
billoneil: I was trying to use the `Seq.take*` but it closed the stream after the predicate matched and we need it to stay open. I will look into a better way to refactor this.

lukaseder: I meant to say: Perhaps `Seq.take` should use this new iterator. In that case, we should open up several tickets and do things one step at a time...

src/main/java/org/jooq/lambda/Seq.java
billoneil: I was also thinking an enum would be the appropriate solution for inclusion / exclusion. It makes sense to work for all four.

>2. The beginning chunk ((1, 1), (2), (2, 3), (4))

Should we assume the beginning chunk starts true? or should this actually return `((2), (2, 3), (4))` skipping anything before the first chunk.

> 4. No chunk

Would this be excluding the matched element? `((1, 1), (), (3))`

src/test/java/org/jooq/lambda/IteratorsTest.java
billoneil: I think they are in the same package so it should have been package-private. The first implementation was must have missed it when I cleaned it up.

>2. I'm not too fan of testing internals with unit tests. The internals will be tested transitively by testing the chunked API. With tests on internals, refactoring those internals will be much harder...

👍 totally agree wasn't sure of your preference so opted for more testing

billoneil:
I will first look into making the TakeIterator and reusing it, then come back to this after. I'll make a new issue and PR once it's ready.

src/main/java/org/jooq/lambda/Seq.java
lukaseder: > Should we assume the beginning chunk starts true? or should this actually return ((2), (2, 3), (4)) skipping anything before the first chunk.

Egh... :) OK, so in the no chunk case, we get:

- 1 chunk for 0 true evaluations of the predicate
- 2 chunks for 1 true evaluation of the predicate
- 3 chunks for 2 true evaluations of the predicate

For example:

```java
// ((1, 1, 2, 2, 3, 4))
Seq.of(1, 1, 2, 2, 3, 4).chunked(i -> false);

// ((),(),(),(),(),(),())
Seq.of(1, 1, 2, 2, 3, 4).chunked(i -> true);
```

I think that options 1-3 should adhere to this logic. My example 1 was wrong, there should be a trailing empty chunk. Will fix the comment.

lukaseder: Hence:

```java
// ((1),(1),(2),(2),(3),(4),())
Seq.of(1, 1, 2, 2, 3, 4).chunked(i -> true, BEFORE);
// ((),(1),(1),(2),(2),(3),(4))
Seq.of(1, 1, 2, 2, 3, 4).chunked(i -> true, AFTER);
// ((1),(1, 1),(1, 2),(2, 2),(2, 3),(3, 4),(4))
Seq.of(1, 1, 2, 2, 3, 4).chunked(i -> true, BOTH);
```

Still a bit undecided about the `BOTH` part, although it could be useful and it would be consistent.

src/test/java/org/jooq/lambda/IteratorsTest.java
lukaseder: I'm all in for more testing, but high level tests will never break because the API won't change again. Low level tests are a bit of a maintenance burden.

lukaseder:
> I will first look into making the TakeIterator and reusing it, then come back to this after. I'll make a new issue and PR once it's ready.

Great, looking forward to this. Note: If it's too complex, we don't have to do it. I just thought if you do create new classes and API, it would be nice if they're reusable. Otherwise, I might prefer inlining them where they're really needed.

src/main/java/org/jooq/lambda/Seq.java
billoneil: We could always ignore `BOTH` for now and easily add it later if requested since it will be an enum.

lukaseder: Yes, let's do that.

lukaseder:
Btw, there's also some work by @tlinkowski pending review. He's implemented a `SeqBuffer`. It goes in a similar direction as this `TakeIterator`: https://github.com/jOOQ/jOOL/issues/305

Maybe, have a look at that first, prior to delving into a refactoring...?

billoneil:
Hmm the `SeqBuffer` doesn't seem like it would work on an infinite sized `Seq` the iterator would. I'll explore some more over the weekend.

billoneil:
My original use case for this would be the ability to Stream something like a very large file in S3 to a Seq, lazily batch N records, batch process N records, stream to a new S3 file. Ideally it should work on any sized file. So I have been working with the assumption of infinite sized streams.

tlinkowski:
I think that for this (#296) and for `Seq.duplicate` to work optimally, `SeqBuffer` would need to support the following things:
- `SeqBuffer.shutdown()` or some other method - after calling this method no more calls to `seq()` would be legal (similarly to Java's `ExecutorService.submit()`)
- field `SeqBuffer.activeSpliterators` inside `SeqBuffer` that would store all created `BufferSpliterators` that haven't yet reached the end of iteration
- field `SeqBuffer.startIndex` that would store the index corresponding to the first element in `List<T> buffer`
- then, upon each call to `SeqBuffer.BufferSpliterator.tryAdvance()`, `SeqBuffer` would be notified by the `BufferSpliterator`, and if this `SeqBuffer` has been shut down, it would:
-- find minimal `nextIndex` among all `activeSpliterators`
-- remove all redundant elements from `List<T> buffer`
-- and use this minimal index as a new `startIndex`
- additionally, `List<T> buffer` could be converted to `LinkedList` upon call to `shutdown()` to improve removal performance

It's not trivial but it seems doable. What do you think, @billoneil, @lukaseder?

billoneil:
I see two use cases when splitting a `Seq`

1. You know the bounds and want to be able to operate on both returned `Seq`'s
 - It would be totally reasonable to buffer the entire Seq (`SeqBuffer`) into memory and have logic so that both returned `Seq`'s can be read concurrently (even if they are backed by a single `Seq`
2. You don't know the bounds and are splitting for some type of chunking.
- In this case You would be forced to read both `Seq`'s sequentially. If you try to read the second before the first is fully consumed you can either throw an error or forward the first `Seq` all the way to completion.

They are both very valid but different use cases. I wouldn't know the best way to portray that in an API since they have different contracts. Maybe allow some flags like `Seq.buffered` vs `Seq.lazy` as a second argument or different methods to handle different implementations? Maybe the API just picks one way always.

One of the reasons I liked the iterator approach is because it's lazy by default and if you want the values in memory you can always call `.toList()` on either or both `Seq`. The issue is what happens if you call `.toList()` on the second before the first?

billoneil:
I see two use cases when splitting a `Seq`

1. You know the bounds and want to be able to operate on both returned `Seq`'s
 - It would be totally reasonable to buffer the entire Seq (`SeqBuffer`) into memory and have logic so that both returned `Seq`'s can be read concurrently (even if they are backed by a single `Seq`
2. You don't know the bounds and are splitting for some type of chunking.
- In this case You would be forced to read both `Seq`'s sequentially. If you try to read the second before the first is fully consumed you can either throw an error or forward the first `Seq` all the way to completion.

They are both very valid but different use cases. I wouldn't know the best way to portray that in an API since they have different contracts. Maybe allow some flags like `Seq.buffered` vs `Seq.lazy` as a second argument or different methods to handle different implementations? Maybe the API just picks one way always.

One of the reasons I liked the iterator approach is because it's lazy by default and if you want the values in memory you can always call `.toList()` on either or both `Seq`. The issue is what happens if you call `.toList()` on the second before the first?

This also just reminded me of something when I first brought this up.

Maybe the `chunked` method should return `Seq<List<E>>` because what happens if someone does the following.

```java
Seq<Seq<Integer>> seqs = ...
seqs.map(s -> seq.limit(5))
```

This would mess up sequential streams unles sunder the hood it ignored the limit and traversed the whole iterator.

billoneil:
I'm going to hold off until this is discussed a bit more.

billoneil:
@lukaseder @tlinkowski Either of you have any thoughts here?

tlinkowski:
@billoneil Well, I'm under the impression that the modified `SeqBuffer` that I described above (supporting `shutdown` method) would absolutely suffice to implement what you described under point no. 2 (BTW. you posted twice almost the same thing - you could delete your first post).

Even more, the advantage of implementing `chunked` using such a modified `SeqBuffer` is that you can read the resulting `Seq`s in any order you like! The only penalty for reading them in a "strange" order are potential performance problems (e.g. if you start reading "from the end", `SeqBuffer` will have to buffer the *entire* source `Seq` into memory). However, if you read these returned "chunk" `Seq`s in order, `SeqBuffer` won't actually have to buffer **anything**! This is because `SeqBuffer` will know (because it will have been shut down) that the item you've just read from the source `Seq` won't be accessible from any other `Seq`. In such scenario, of course, anyone can call anything on the resulting `Seq`s (e.g. `seq.limit(5)` that you mentioned) and it doesn't break anything - all calls just forward to `SeqBuffer.BufferSpliterator`, which (together with `SeqBuffer`) takes care of what needs to be buffered and what can be discarded

One more thought about the modified `SeqBuffer` implementation - to construct such `chunked` `Seq`s we couldn't just use `SeqBuffer.seq().skip(n)` because this would yield `SeqBuffer.BufferSpliterator`s with `nextIndex = 0`, which in turn would lead to a situation where `SeqBuffer` cannot discard anything until the `Spliterator`s of all the `Seq`s have been advanced. So we'd need to add some extra method (like `SeqBuffer.seqFrom(n)`) that would create a `SeqBuffer.BufferSpliterator` with `nextIndex = n`.

billoneil:
> The only penalty for reading them in a "strange" order are potential performance problems (e.g. if you start reading "from the end", SeqBuffer will have to buffer the entire source Seq into memory)

This sounds reasonable as long as it's documented.

> However, if you read these returned "chunk" Seqs in order, SeqBuffer won't actually have to buffer anything!

I might have missed this part when I read the source the first time. This does indeed seem like it can cover all use cases.

> In such scenario, of course, anyone can call anything on the resulting Seqs (e.g. seq.limit(5) that you mentioned) and it doesn't break anything - all calls just forward to SeqBuffer.BufferSpliterator, which (together with SeqBuffer) takes care of what needs to be buffered and what can be discarded

I think this would also work and it just needs to be documented since it could be doing extra work under the hood than expected.

tlinkowski:
@billoneil You haven't missed anything in the source code :) I might have been imprecise but please note that in [my most recent comment](https://github.com/jOOQ/jOOL/pull/320#issuecomment-320871428) I was referring to [my previous comment](https://github.com/jOOQ/jOOL/pull/320#issuecomment-318704533) that contains only a proposal of adapting `SeqBuffer` to do what is needed to efficiently implement `chunked`. But it's just an idea - there's no implementation yet, although I'd gladly provide one if @lukaseder confirms that he likes the idea and is willing to have this implemented.

lukaseder:

Issue #350, opened 2019-06-19
Add support for unbounded ranges


lukaseder:
The `org.jooq.lambda.tuple.Range` type currently supports only bounded ranges. It would be nice to support also unbounded ranges, where `null` corresponds to "infinity".

For example:

```java
// From 1 to infinity
new Range(1, null);

// From -infinity to 1
new Range(null, 1);
```

tlinkowski:
Just for reference (in case you wanted to further improve your ranges), [Guava](https://github.com/google/guava/wiki/RangesExplained) has quite advanced range support (including e.g. open/closed bounds).

Artus2b:
@tlinkowski Indeed I found the Guava and Apache Common Ranges implementations, while I was writing the algorithm of `overlaps(Range a, Range b)` that was a bit cumbersome.
Are you saying those libs nullify the usefulness of the issue? Or that we should use Guava's implementation in the proposed fix?

tlinkowski:
> Are you saying those libs nullify the usefulness of the issue? Or that we should use Guava's implementation in the proposed fix?

@Artus2b I don't know which is best here. In your PR, you mentioned how it works in PostgreSQL, so aligning to its Range Types (independently of other libs) in certainly one option.

To sum up, I'm just saying it's good to be *aware* of other solutions (thx for mentioning Apache Commons - forgot about that) so that an informed decision can be made.

lukaseder:
> Are you saying those libs nullify the usefulness of the issue?

Everything jOOλ does is available in one way or another through a different means. After all, we're all working with turing complete languages, so everything is possible, including NIH, home grown solutions to every single problem :)

Given that we already have a `Range` type, it's definitely worth exploring what else it can offer, and I do like this idea, regardless of what Guava does.

lukaseder:
I think that all changes from your PR are now addressed:

- Documentation of inclusivity of current range implementation
- (Partially) Unbounded ranges
- `Range.contains()`

Thanks again for your suggestion and PR.

PR #351, opened 2019-06-20
[#350] allow creation of unbounded Ranges

Artus2b:
The class `org.jooq.lambda.tuple.Range` does not support nullable bounds.

For example in PostgreSQL a lower range bound of value `null` means `-infinity`.

See the [Range Types documentation](https://www.postgresql.org/docs/11/rangetypes.html#RANGETYPES-INFINITE):
> The lower bound of a range can be omitted, meaning that all points less than the upper bound are included in the range.

E.g. when testing that the range `(,3)` contains the value `-12`:
```
# SELECT '(,3)'::INT4RANGE @> -12;
 ?column?
----------
 t
(1 row)
```

## What's left to do in this PR
I did not update the method `Tuple2.intersect` to handle null bounds, because it's a bit complicated, and I want to make sure that people support this PR before making this last effort :smile:.

## Other possible improvements
- support bound inclusiveness or exclusiveness (in documentation "[Inclusive and Exclusive Bounds](https://www.postgresql.org/docs/11/rangetypes.html#RANGETYPES-INCLUSIVITY)")
- port more functions than `intersect` and `overlaps` (in documentation "[Range Functions and Operators](https://www.postgresql.org/docs/11/functions-range.html)")

## Usage example

For PostgreSQL type `tstzrange`:

```java
public class TsTzRangeBinding implements Binding<Object, Range<Instant>> {

    public class TsTzRangeConverter implements Converter<Object, Range<Instant>> {

        private static final Pattern PATTERN = Pattern.compile("\\[(.*?),(.*?)\\)");

        @Override
        public Range<Instant> from(Object t) {
            if (t == null) {
                return null;
            }

            Matcher m = PATTERN.matcher("" + t);
            if (m.find()) {
                return new Range<>(Instant.parse(m.group(1)), Instant.parse(m.group(2)));
            }

            throw new IllegalArgumentException("Unsupported range : " + t);
        }

        @Override
        public Object to(Range<Instant> u) {
            return u == null ? null : "[" + emptyIfNull(u.v1) + "," + emptyIfNull(u.v2) + ")";
        }

        private String emptyIfNull(Instant i) {
            return i == null ? "" : i.toString();
        }

        @Override
        public Class<Object> fromType() {
            return Object.class;
        }

        @SuppressWarnings({"unchecked", "rawtypes"})
        @Override
        public Class<Range<Instant>> toType() {
            return (Class) Range.class;
        }

    }

    @Override
    public Converter<Object, Range<Instant>> converter() {
        return new TsTzRangeConverter();
    }

    @Override
    public void sql(BindingSQLContext<Range<Instant>> ctx) {
        ctx.render()
                .visit(DSL.val(ctx.convert(converter()).value()))
                .sql("::TSTZRANGE");
    }

    // ...
}

```
... And register the binding:
```java
ForcedType tstzrange = new ForcedType()
                .withUserType("org.jooq.lambda.tuple.Range<Instant>")
                .withBinding("org.artus.jooq.binding.TsTzRangeBinding")
                .withTypes("tstzrange");
```

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
@@ -35,18 +35,23 @@ public class Range<T extends Comparable<T>> extends Tuple2<T, T> {
     }
 
     private static <T extends Comparable<T>> Tuple2<T, T> r(T t1, T t2) {
-        return t1.compareTo(t2) <= 0 ? new Tuple2<>(t1, t2) : new Tuple2<>(t2, t1);
+        if (t1 != null && t2 != null)
+            return t1.compareTo(t2) <= 0 ? new Tuple2<>(t1, t2) : new Tuple2<>(t2, t1);
+        else
+            return new Tuple2<>(t1, t2);
     }
 
     /**
-     * Whether two ranges overlap.
+     * Whether two ranges overlap. Assumes inclusiveness.
      * <p>
      * <code><pre>
      * // true
      * range(1, 3).overlaps(range(2, 4))
+     * range(1, null).overlaps(range(null, 1))
      *
      * // false
      * range(1, 3).overlaps(range(5, 8))
+     * range(null, 3).overlaps(range(5, null))
      * </pre></code>
      */
     public boolean overlaps(Tuple2<T, T> other) {
@@ -96,5 +101,39 @@ public class Range<T extends Comparable<T>> extends Tuple2<T, T> {
      */
     public Optional<Range<T>> intersect(T t1, T t2) {
         return intersect(new Range<>(t1, t2));
+    }
+
+    /**
+     * Whether the given value is included in the range. Assumes range inclusiveness.
+     * <p>
+     * <code><pre>
+     * // true
+     * range(1, 3).contains(1)
+     * range(2, null).contains(5)
+     *
+     * // false
+     * range(1, 3).contains(4)
+     * range(2, null).contains(1)
+     * </pre></code>
+     */
+    public boolean contains(T t) {
+        return Tuple2.contains(this, t);
+    }
+
+    /**
+     * Whether the given right tuple is included in the left tuple.
+     * <p>
+     * <code><pre>
+     * // true
+     * range(1, 3).contains(range(1, 2))
+     * range(2, null).contains(range(3, 5))
+     *
+     * // false
+     * range(1, 3).contains(range(null, 2))
+     * range(2, null).contains(range(1, 3))
+     * </pre></code>
+     */
+    public boolean contains(Tuple2<T, T> other) {
+        return Tuple2.contains(this, other);
     }
 }
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
@@ -15,21 +15,13 @@
  */
 package org.jooq.lambda.tuple;
 
+import org.jooq.lambda.Seq;
+
 import java.io.Serializable;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
-
-import org.jooq.lambda.Seq;
-import org.jooq.lambda.function.Function1;
-import org.jooq.lambda.function.Function2;
 
 /**
  * A tuple of degree 2.
@@ -237,19 +229,39 @@ public class Tuple2<T1, T2> implements Tuple, Comparable<Tuple2<T1, T2>>, Serial
     }
 
     /**
-     * Whether two tuples overlap.
+     * Whether two tuples overlap. Assumes inclusiveness.
      * <p>
      * <code><pre>
      * // true
      * range(1, 3).overlaps(range(2, 4))
+     * range(1, null).overlaps(range(null, 1))
      *
      * // false
      * range(1, 3).overlaps(range(5, 8))
+     * range(null, 3).overlaps(range(5, null))
      * </pre></code>
      */
     public static final <T extends Comparable<? super T>> boolean overlaps(Tuple2<T, T> left, Tuple2<T, T> right) {
-        return left.v1.compareTo(right.v2) <= 0
-            && left.v2.compareTo(right.v1) >= 0;
+        if (left == null || right == null) {
+            return false;
+        }
+
+        if (left.infinite() || right.infinite()) {
+            return true;
+        }
+        if (left.finite() && right.finite()) {
+            return left.v1.compareTo(right.v2) <= 0
+                    && left.v2.compareTo(right.v1) >= 0;
+        }
+
+        if (!left.upperInfinite() && !right.lowerInfinite()) {
+            return left.v2.compareTo(right.v1) >= 0;
+        }
+        if (!left.lowerInfinite() && !right.upperInfinite()) {
+            return left.v1.compareTo(right.v2) <= 0;
+        }
+        // two remaining positive cases: either both have upper or lower bound infinite
+        return true;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
```diff
@@ -35,18 +35,23 @@ public Range(Tuple2<T, T> tuple) {
     }
 
     private static <T extends Comparable<T>> Tuple2<T, T> r(T t1, T t2) {
-        return t1.compareTo(t2) <= 0 ? new Tuple2<>(t1, t2) : new Tuple2<>(t2, t1);
+        if (t1 != null && t2 != null)
+            return t1.compareTo(t2) <= 0 ? new Tuple2<>(t1, t2) : new Tuple2<>(t2, t1);
+        else
+            return new Tuple2<>(t1, t2);
```
lukaseder: The fact that this is necessary raises more questions: Is swapping input bounds still a valid thing in the presence of possibly unbounded ranges? I think it isn't. But if we want to stay backwards compatible, then this logic is definitely correct...

I've created an issue for this, to keep this in mind: #352

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
```diff
@@ -97,4 +102,38 @@ public boolean overlaps(T t1, T t2) {
...
+     * range(2, null).contains(range(3, 5))
+     *
+     * // false
+     * range(1, 3).contains(range(null, 2))
+     * range(2, null).contains(range(1, 3))
+     * </pre></code>
+     */
+    public boolean contains(Tuple2<T, T> other) {
+        return Tuple2.contains(this, other);
+    }
```
lukaseder: This is an entirely unrelated feature, which should be tracked and discussed in a separate issue.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -264,15 +276,76 @@ public final Tuple0 skip2() {
...
+        return v2 == null;
+    }
+
+    public final <T extends Comparable<? super T>> boolean finite() {
+        return !lowerInfinite() && !upperInfinite();
+    }
+
+    public final <T extends Comparable<? super T>> boolean infinite() {
+        return lowerInfinite() && upperInfinite();
+    }
```
lukaseder: All of this added API:

- Should be added to `Range` only, not to `Tuple2`. A `Tuple2` has no notion of one of its attributes being "infinite", just because it happens to be `null`.
- In my opinion doesn't deserve to be public API anyway.

lukaseder:
Thanks a lot for your suggestions. There are too many review findings for this to be merged as it is. But I will take the following bits from your commits and integrate them in slightly changed ways:

- The tests
- The contains method (but implemented differently: https://github.com/jOOQ/jOOL/issues/354)
- Parts of the implementation

Thanks again for taking the time to suggest an implementation.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
lukaseder: I realise that this implementation was put into `Tuple2` because `overlaps()` is there. Which might have been a mistake...

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -15,22 +15,14 @@
...
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
-import org.jooq.lambda.Seq;
-import org.jooq.lambda.function.Function1;
-import org.jooq.lambda.function.Function2;
```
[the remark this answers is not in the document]
Artus2b: my IDE's is set to do it automatically sorry :grimacing:

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
lukaseder: We'll deprecate these methods on `Tuple2`: https://github.com/jOOQ/jOOL/issues/355

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -237,19 +229,39 @@ public final Tuple0 skip2() {
...
+        }
+
+        if (!left.upperInfinite() && !right.lowerInfinite()) {
+            return left.v2.compareTo(right.v1) >= 0;
+        }
+        if (!left.lowerInfinite() && !right.upperInfinite()) {
+            return left.v1.compareTo(right.v2) <= 0;
+        }
+        // two remaining positive cases: either both have upper or lower bound infinite
+        return true;
```
lukaseder: This seems much more concise:

```java
    public boolean overlaps(Range<T> other) {
        return (v1 == null
            ||  other.v2 == null
            ||  v1.compareTo(other.v2) <= 0)
            && (v2 == null
            ||  other.v1 == null
            ||  v2.compareTo(other.v1) >= 0);
    }
```

Artus2b: Yes but at the cost of readability.
But readability is a bit subjective of course.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
Artus2b: yes I found that a bit weird to find them in `Tuple2`, I'm glad to read you say it's a mistake.
I've added my method next to the others as to disturb the code base the less possible, and to not make architectural choices outside the scope of the PR.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -264,15 +276,76 @@ public final Tuple0 skip2() {
...
             return Optional.of(new Tuple2<>(
-                left.v1.compareTo(right.v1) >= 0 ? left.v1 : right.v1,
-                left.v2.compareTo(right.v2) <= 0 ? left.v2 : right.v2
+                    left.v1.compareTo(right.v1) >= 0 ? left.v1 : right.v1,
+                    left.v2.compareTo(right.v2) <= 0 ? left.v2 : right.v2
             ));
-        else
+        } else
             return Optional.empty();
     }
```
lukaseder: For the record: This is how `intersect` could be implemented:

```java
    public Optional<Range<T>> intersect(Range<T> other) {
        if (overlaps(other))
            return Optional.of(new Range<>(
                v1 == null
                    ? other.v1
                    : other.v1 == null
                    ? v1
                    : v1.compareTo(other.v1) >= 0
                    ? v1
                    : other.v1,
                v2 == null
                    ? other.v2
                    : other.v2 == null
                    ? v2
                    : v2.compareTo(other.v2) <= 0
                    ? v2
                    : other.v2
            ));
        else
            return Optional.empty();
    }
```

Artus2b:
hey @lukaseder thanks for your time and comments/

So am I to submit corrected changes in another PR? or you are 'taking it from here' and use what's here to complete the change?
Is it the usual process on this project? I would have imagined that based on the comments received I could add 'fix' commits to this PR.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -264,15 +276,76 @@ public final Tuple0 skip2() {
...
+     * </pre></code>
+     */
+    @SafeVarargs
+    public static final <T extends Comparable<? super T>> boolean contains(Tuple2<T, T> tuple2, T... vals) {
+        return Arrays.stream(vals).map(v -> {
+            if (tuple2.v1 != null && tuple2.v1.compareTo(v) > 0)
+                return false; // value is below the lower bound
+            if (tuple2.v2 != null && tuple2.v2.compareTo(v) < 0)
+                return false; // value is above the upper bound
+            return true;
```
lukaseder: To stay on the safe side, each `val` should be checked for `null` as well, given that in this context, `null` have a special meaning, they might appear as well here.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -264,15 +276,76 @@ public final Tuple0 skip2() {
...
+        }).reduce(Boolean::logicalAnd).orElse(false);
```
lukaseder: I'm not sure we need a multi-val `contains()` method, but in case we do, it would be much better to "fail" early, instead of reducing the entire stream. In imperative logic, this would be much simpler to code (and faster, given we'll need less stack frames and have more inlining potential). But if you want to use a stream, I think you could use:

```java
}).filter(v -> !v).findAny().orElse(true);
```

Artus2b: That's a good advice thank you.
Stream wasn't a requirement here, it was just used in the other method `Range#contains(Range other)` so I could call this `contains` method with both bound instead of calling it twice.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
```diff
@@ -264,15 +276,76 @@ public final Tuple0 skip2() {
...
+     * contains(range(2, null), range(3, 5))
+     *
+     * // false
+     * contains(range(1, 3), range(null, 2))
+     * contains(range(2, null), range(1, 3))
+     * </pre></code>
+     */
+    public static final <T extends Comparable<? super T>> boolean contains(Tuple2<T, T> left, Tuple2<T, T> right) {
+        return contains(left, right.v1, right.v2);
+    }
```
lukaseder: This implementation is incorrect when the right side range is (partially) unbounded. A correct implementation would be this:

```java
    public boolean contains(Range<T> other) {
        return (other.v1 == null && v1 == null || contains(other.v1))
            && (other.v2 == null && v2 == null || contains(other.v2));
    }
```

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
lukaseder: Sure, we can use subjectivity here.

I find your version much less readable, because it takes me much longer to digest. For example, why is `left.infinite() || right.infinite()` a special case worth short circuiting? Why is the `left.finite() && right.finite()` version short circuited right after? Why do we check for infinity several times?

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
Artus2b: yes that's an error :disappointed:

lukaseder:
> So am I to submit corrected changes in another PR? or you are 'taking it from here' and use what's here to complete the change?

I'm taking it from here.

> Is it the usual process on this project? I would have imagined that based on the comments received I could add 'fix' commits to this PR.

That's what I thought at first, until I saw how much needed fixing, at which point the ping pong of getting PRs, reviewing, getting PRs, reviewing takes much more of both our time than just doing it myself.

> Edit: Oh I've just seen your change on the issue. all good then

Yep :)

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Tuple2.java
lukaseder: Indeed, nice observation. Over the past years, I've (drastically) changed my mind about the utility of `Stream` usage, especially in library internals. It's still OK to use it in business logic, where it may indeed improve readability over imperative logic (sometimes it doesn't improve it), at a negligible performance overhead, because business logic is much more I/O bound (e.g. databases) than CPU bound. But in libraries, the stream overhead can be a killer.

jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
lukaseder: Makes sense, thanks.

Issue #277, opened 2016-12-15
Add Seq.flatFold{Left,Right}


thomas001:
When using foldLeft, it may be useful to abort iteration, and not traverse the whole sequence. Using a Optional<T> return value might just do the trick, for example:
```java
public static <S,U> Optional<S> flatFoldLeft(Iterable<? extends U> list, 
                                             Optional<S> seed, 
                                             BiFunction<S,U,Optional<S>> func) {
    Iterator<? extends U> iterator = list.iterator();
    while(iterator.hasNext() && seed.isPresent()) {
        U nextVal = iterator.next();
        seed = func.apply(seed.get(), nextVal);
    }
    return seed;
}
```

lukaseder:
Hmm, interesting, thanks for your suggestion. I have two questions:

- Where does the name come from? Is this a commonly used term: "flatFold"?
- What about this instead: `seq.limitUntil(something).foldLeft(seed, function)`?

lukaseder:
I don't see much value in this additional API. I'm happy to discuss requirements / examples / use cases, but I'd like to avoid adding more features eagerly to the API.

PR #366, opened 2020-05-31
[#277] Add Seq.flatFold{Left, Right} to Seq.java

YETSDD:
When using foldLeft or foldRight, there will be a situation that iterations need to be abort.  We add the limited flatfold to deal with this situation. For example, if we want to get a least sum more than 10 from a seq (1, 2, 3, 4, 5, 6) we can use this function to get 15


lukaseder:
There have been numerous PRs without prior discussion in recent days.

While I appreciate the effort, I really like to discuss things first before going ahead with specific implementations.

In this case, for example, I don't fully undrestand the choice of name nor the requirement itself, see also #277 

Thanks for discussing ideas first.

Issue #333, opened 2018-03-02
Adapt all Seq.crossJoin() methods to make use of SeqBuffer


tlinkowski:
This is related to #305, where the following `Seq` methods were declared eligible for reimplementation using `SeqBuffer`:
- `crossSelfJoin()` (done in #305)
- `inner(Self)Join()` (done in #305)
- `leftOuter(Self)Join()` (done in #305)
- `crossJoin()` (not done because it needs to be generated using `jOOQ-tools`)

As a consequence, the 0.9.13 release has some `SeqBuffer`-eligible methods that are implemented using `SeqBuffer` and some that are not.

### Expected behavior and actual behavior:
Expected: All `Seq.crossJoin()` variants are implemented using `SeqBuffer`.
Actual: All `Seq.crossJoin()` variants are implements using `Seq.toList()`.

### Versions:

- jOOλ: 0.9.13

tlinkowski:
I've noticed that this issue has not been addressed at all (it's not even labeled) so I hereby make a "ping" :)

lukaseder:
Thanks a lot for your ping.

I'm sorry for the delays in these matters. Whether or not an alternative implementation for a variety of jOOλ API should be implemented is a rather time consuming task, time which I currently - very unfortunately - cannot allocate. This doesn't mean that jOOλ will not be improved, but it does mean that the cost / benefit ratio for complex changes like this one is currently not very good. There are lower hanging fruit which have higher chances of making it in the next releases.

PR #367, opened 2020-05-31
[#333] Adapt all Seq.crossJoin() methods to make use of SeqBuffer

YETSDD:
As a consequence, the 0.9.13 release has some SeqBuffer-eligible methods that are implemented using SeqBuffer and some that are not. I change the crossjoin method by implementing it by SeqBuffer() rather than toList()


lukaseder:
Thanks for your suggestion. I rest my case of my comment here: https://github.com/jOOQ/jOOL/issues/333#issuecomment-440216973

The big effort in this area is not to actually implement things, but to take time to think about these things, review them, plan them, etc.

I currently do not want to allocate time to review the eligibility of `SeqBuffer` in various places inside of jOOλ

Issue #275, opened 2016-12-14
Add Seq.fullOuterJoin()


lukaseder:

jmax01:
Below is a rough example of a workaround:

```java 

import static java.util.Optional.*;
import static org.jooq.lambda.Seq.*;

import java.util.List;
import java.util.stream.Collectors;
import java.util.stream.Stream;

import org.jooq.lambda.Seq;
import org.jooq.lambda.tuple.Tuple2;
import org.junit.Test;

import lombok.Value;

public class JoolTest {

    @Test
    public void matchPositions() {

        List<Position> client1Positions = Stream.of(Position.of("ticker1", "client1", 0.0),
                Position.of("ticker2", "client1", 0.0), Position.of("ticker4", "client1", 0.0))
            .collect(Collectors.toList());

        List<Position> client2Postions = Stream.of(Position.of("ticker1", "client2", 0.0),
                Position.of("ticker3", "client2", 0.0), Position.of("ticker4", "client2", 0.0))
            .collect(Collectors.toList());

        Seq<Tuple2<Position, Position>> leftJoinClient1ToClient2 = seq(client1Positions).leftOuterJoin(client2Postions,
                Position::sameTicker);

        Seq<Tuple2<Position, Position>> rightJoinClient1ToClient2 = seq(client1Positions)
            .rightOuterJoin(client2Postions, Position::sameTicker);

        // Expected Output:
        //
        // (JoolTest.Position(ticker=ticker1, client=client1, notional=0.0), JoolTest.Position(ticker=ticker1,
        // client=client2, notional=0.0))
        // (JoolTest.Position(ticker=ticker2, client=client1, notional=0.0), null)
        // (JoolTest.Position(ticker=ticker4, client=client1, notional=0.0), JoolTest.Position(ticker=ticker4,
        // client=client2, notional=0.0))
        // (null, JoolTest.Position(ticker=ticker3, client=client2, notional=0.0))
        leftJoinClient1ToClient2.concat(rightJoinClient1ToClient2)
            .distinct()
            .stream()
            .forEach(System.out::println);

    }

    @Value(staticConstructor = "of")
    static class Position {

        String ticker;

        String client;

        double notional;

        public boolean sameTicker(Position that) {

            return ofNullable(that).filter(t -> getTicker().equals(t.getTicker()))
                .isPresent();

        }
    }
```

lukaseder:
Thanks, @jmax01

PR #368, opened 2020-05-31
[#275]Implement Seq.fullOuterJoin in Seq.java

YETSDD:
We add the fullOuterJoin and fullOuterSelfJoin by combining leftOuterJoin and RightOuterJoin


lukaseder:
Thanks for your suggestion. The implementation is incorrect, which is why I'm rejecting this PR.

Issue #276, opened 2016-12-14
Add Seq.partitionedOuterJoin()


lukaseder:
The actual method name may vary

PR #369, opened 2020-05-31
[#276] Implement Seq.partitionOuterJoin in Seq.java

YETSDD:
partitionOuterJoin here is  a relationship bewteen tuples and it has the same effect as partitionOuterJoin in SQL


lukaseder:
There have been numerous PRs without prior discussion in recent days.

While I appreciate the effort, I really like to discuss things first before going ahead with specific implementations.

Thanks for discussing ideas first.

Issue #359, opened 2019-09-26
Add Agg.taking() and Agg.dropping() collectors


lukaseder:
The `Seq.take()` and `Seq.drop()` operations are useful to limit data set sizes, just like SQL's `OFFSET .. FETCH`. Similar functionality would be equally useful in the form of a `Collector`. We already have `Agg.first()` and `Agg.last()`, which are special cases.

varunu28:
@lukaseder I will like to work on this feature. Can you please assign it to me? Thanks

ZeroZill:
Maybe this should work?
```Java
public static <T> Collector<T, ?, Seq<T>> taking(final long takeNumber) {
        return Collector.of(
                (Supplier<LinkedList<T>>) LinkedList::new,
                (l, v) -> {
                    if (l.size() < takeNumber) {
                        l.add(v);
                    }
                },
                (l1, l2) -> {
                  l1.addAll(l2);
                  return l1;
                },
                l -> Seq.seq(l.stream())
        );
    }

public static <T> Collector<T, ?, Seq<T>> dropping(final long skipNumber) {
        long[] index = {0};

        return Collector.of(
                (Supplier<LinkedList<T>>) LinkedList::new,
                (l, v) -> {
                    if (index[0] >= skipNumber) {
                        l.add(v);
                    }
                    index[0]++;
                },
                (l1, l2) -> {
                    l1.addAll(l2);
                    return l1;
                },
                l -> Seq.seq(l.stream())
        );
    }
```

lukaseder:
Thanks for your suggestions, folks.

PR #370, opened 2020-05-31
[#359]Add Agg.taking and Agg.dropping to Agg.java

TRebirthC:
Agg.taking is to get the first n elements in a collector. Agg.dropping is to get the last n elements in a collector.

### DIFF
--- a/jOOL/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL/src/main/java/org/jooq/lambda/Agg.java
@@ -83,6 +83,57 @@ public class Agg {
     }
 
     /**
+     * Get a {@link Collector} that takes the first <code>n</code> elements from a collection.
+     * <p>The final type of this {@link Collector} is {@link Seq} type.
+     * @param n     The number of elements to be taken
+     * @param <T>   The type of element
+     * @return      A {@link Seq} of type <code>T</code>
+     */
+    public static <T> Collector<T, ?, Seq<T>> taking(long n) {
+        return Collector.of(
+                (Supplier<LinkedList<T>>) LinkedList::new,
+                (l, v) -> {
+                    if (l.size() < n) {
+                        l.add(v);
+                    }
+                },
+                (l1, l2) -> {
+                    l1.addAll(l2);
+                    return l1;
+                },
+                l -> Seq.seq(l.stream())
+        );
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -82,6 +82,57 @@
...
+    /**
+     * Get a {@link Collector} that takes the first <code>n</code> elements from a collection.
+     * <p>The final type of this {@link Collector} is {@link Seq} type.
+     * @param n     The number of elements to be taken
+     * @param <T>   The type of element
+     * @return      A {@link Seq} of type <code>T</code>
+     */
+    public static <T> Collector<T, ?, Seq<T>> taking(long n) {
+        return Collector.of(
+                (Supplier<LinkedList<T>>) LinkedList::new,
```
lukaseder, line 94: I don't think we should be using `LinkedList` here. An `ArrayList` seems to be a better choice.

jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -82,6 +82,57 @@
...
+    /**
+     * Get a {@link Collector} that skip the first <code>n</code> elements of a collection.
+     * <p>This method will skip the first <code>n</code> elements, and start from the (n+1)th element.
+     * The final type of this {@link Collector} is {@link Seq} type.
+     * @param n     The number of elements to be skipped
+     * @param <T>   The type of element
+     * @return      A {@link Seq} of type <code>T</code>
+     */
+    public static <T> Collector<T, ?, Seq<T>> dropping(long n) {
+        long[] index = {0};
```
lukaseder, line 117: I'd prefer a `Collector` that doesn't use any capturing state like this. The `Collector.supplier()` should be able to produce an `Accumulator` type.

jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -82,6 +82,57 @@
...
+
+        return Collector.of(
+                (Supplier<LinkedList<T>>) LinkedList::new,
+                (l, v) -> {
+                    if (index[0] >= n) {
+                        l.add(v);
+                    }
+                    index[0]++;
```
lukaseder, line 125: Once we've reached the threshold, we can stop counting

lukaseder:
I'm merging this, fixing things directly myself, as suggested.

Issue #217, opened 2016-03-17
Add Tuple.function[N](Function<? super Tuple[N]<T1, ..., TN>, ? extends R>


lukaseder:
This is the inverse operation for #214

```java
interface Tuple {
    static Consumer2<T1, T2> consumer2(Consumer<? super Tuple2<T1, T2>> consumer);
    static <R> Function2<T1, T2, R> function2(Function<? super Tuple2<T1, T2>, ? extends R> function);
}
```

A lot of methods would be generated this way. Perhaps we should wait with the implementation to see if this is really needed.

----
See also: https://github.com/jOOQ/jOOL/issues/214#issuecomment-197309302

Issue #214, opened 2016-03-16
Add Consumer1<Tuple[N]<T1, ..., TN>> Tuple.consumer(Consumer[N]<T1, ..., TN>) and Function1<Tuple[N]<T1, ..., TN>> Tuple.function(Function[N]<T1, ..., TN>)


lukens:
Does anything like the following currently exist in jOOL, or would there be any appetite for adding something like it? I created my own, as I couldn't find anything.

```java
public class Tuple2Lambdas {

    public static <A, B> Predicate<Tuple2<A, B>> predicate(BiPredicate<A, B> predicate) {

        return (tuple) -> predicate.test(tuple.v1, tuple.v2);

    }

    public static <A, B, R> Function<Tuple2<A, B>, R> function(BiFunction<A, B, R> function) {

        return (tuple) -> function.apply(tuple.v1, tuple.v2);

    }

    public static <A, B> Consumer<Tuple2<A, B>> consumer(BiConsumer<A, B> consumer) {

        return (tuple) -> consumer.accept(tuple.v1, tuple.v2);

    }

}
```

This allows the following:

```java
        Seq.seq(tuple2List)
            .filter(predicate(this::validHit))
            .map(function(this::getDisplayMatch))
            .forEach(consumer(this::processResult));
```

or:

```java
        Seq.seq(tuple2List)
            .filter(predicate((hit, property) -> hit.getId().equals("blah")))
            .map(function((hit, property) -> tuple(hit.explanation(), property.getId())))
```

rather than the more long winded:

```java
        Seq.seq(tuple2List)
            .filter(tuple -> validHit(tuple.v1, tuple.v2))
            .map(tuple -> getDisplayMatch(tuple.v1, tuple.v2))
            .forEach(tuple -> processResult(tuple.v1, tuple.v2));
```

or more opaque: 

```java
        Seq.seq(tuple2List)
            .filter((tuple) -> tuple.v2.getId().equals("blah"))
            .map((tuple) -> tuple(tuple.v1.explanation(), tuple.v2.getId())
```

Most times when I've rolled my own solutions to Java's lambda limitations I then find jOOL has beaten me to it, but I couldn't find anything in this case.

lukens:
Something like the following allows it to other `TupleN`s (though consumers are trickier, and would probably require `ConsumerN`s to match the `FunctionN`s [and the predicates are a little hacky]):

```java
public class TupleLambdas {

[...]

PR #371, opened 2020-05-31
[#217]Add Tuple.function[N]

TRebirthC:
This is the inverse operation for [issue:#214].


lukaseder:
Thanks for your suggestion. I'm not in a hurry to add these methods as mentioned in https://github.com/jOOQ/jOOL/issues/217

I'd like to gather feedback on the utility of them first, and then still perhaps decide that we don't need them.

Please discuss your suggestions first before engaging in working on them, there's always a risk of doing unnecessary work (notice also, we have a code generator to generate the API for `1-16` overloads! I hope you didn't do all of this manually?)

Issue #221, opened 2016-03-20
Add Agg.medianAll(), Agg.percentileAll() to return a Seq of multiple values


lukaseder:
See also #220

PR #372, opened 2020-05-31
[#221]Add Agg.medianAll() and Agg.percentileAll()

TRebirthC:
It will return a Seq of multiple values.

### DIFF
--- a/jOOL/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL/src/main/java/org/jooq/lambda/Agg.java
@@ -41,6 +41,10 @@ import org.jooq.lambda.tuple.Tuple2;
  */
 public class Agg {
 
+    private static double half = 0.5;
+    private static double zero = 0.0;
+    private static double one = 1.0;
+
     /**
      * Get a {@link Collector} that filters data passed to downstream collector.
      */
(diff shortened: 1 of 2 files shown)


jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -886,6 +970,126 @@ else if (percentile == 1.0)
...
+    public static <T, U> Collector<T, ?, Seq<T>> percentileAllBy(final double percentile, final Function<? super T, ? extends U> function, final Comparator<? super U> comparator) {
+        if (percentile < 0.0 || percentile > 1.0) {
+            throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");
+        }
+
+        if (percentile == zero) {
+            return minAllBy(function, comparator);
+        } else if (percentile == one) {
+            return maxAllBy(function, comparator);
+        }
```
lukaseder, line 1067: I like this optimisation. We should implement it also for the ordinary `percentile()` calculations: https://github.com/jOOQ/jOOL/issues/376

jOOL/src/test/java/org/jooq/lambda/CollectorTests.java
```diff
@@ -848,6 +849,36 @@ public void testMinMax() {
...
         assertEquals(Optional.of(1), Seq.of(1, 2, 3).collect(maxBy(i -> -i)));
     }
+
+    @Test
+    public void testPercentileAll(){
+        assertEquals(List.of(), Seq.<Integer>of().collect(percentileAll(1.0)).toList());
+        assertEquals(List.of(1), Seq.of(1).collect(percentileAll(0.0)).toList());
+        assertEquals(List.of(3, 3), Seq.of(1, 3, 3, 2, 1, 2).collect(percentileAll(1.0)).toList());
+        assertEquals(List.of(2, 2), Seq.of(1, 1, 2, 2, 3, 3).collect(percentileAll(0.5)).toList());
+        assertEquals(List.of(1, 1), Seq.of(1, 1, 2, 2, 3, 3).collect(percentileAll(0.0)).toList());
```
lukaseder, line 859: We cannot use these list literals yet, because we still support Java 8 as well.

jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -886,6 +970,126 @@ else if (percentile == 1.0)
...
+                        return Seq.empty();
+                    } else if (size == one) {
+                        return Seq.of(l.get(0).v1);
+                    }
+
+                    l.sort(Comparator.comparing(t -> t.v2, comparator));
+                    // x.5 should be rounded down
+                    final int index = (int) -Math.round(-(size * percentile + half)) - 1;
+                    final T element = l.get(index).v1;
+                    return Seq.seq(l.stream().filter(t -> t.v1.equals(element)).map(t -> t.v1));
```
lukaseder, line 1089: It seems to me that in edge cases, this additional `O(N)` traversal is more expensive than a possibly more optimal way to find the "ties" of the result.

In addition to this, I wonder what the best way to implement `percentileAll()` would be. Is it really desireable to find all ties? Or only ties that are located *after* the percentile position?

This question is already answered partially by the behaviour of `maxAllBy()`, which produces *all* ties, regardless of their position with respect to the percentile value (1.0).

I'm undecided. And once more, I wonder whether we should add this functionality, or wait for someone who actually needs it to think about this more thoroughly.

I prefer the latter.

jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -41,6 +41,10 @@
  */
 public class Agg {
 
+    private static double half = 0.5;
+    private static double zero = 0.0;
+    private static double one = 1.0;
```
lukaseder, line 46: These are not really needed, and I prefer not to create such constants (apart from the fact that they aren't constants)

lukaseder:
Thanks for the many suggestions.

As in previous pull requests, there are a few issues with both the implementation and the feature request per se. I would really like to discuss things first before looking at specific implementations. The fact that there is a feature request means that there is an idea worth discussing. It does not necessarily mean that it is an idea worth implementing.

Issue #154, opened 2016-01-02
Overload groupBy() and grouping() to support multiple collectors


lukaseder:
See also #152

lukaseder:
As shown by a recent PR, this leads to an excessive amount of overloads, which I don't want to add (for now), for these operations.

PR #373, opened 2020-05-31
[#154]Overload groupBy() and grouping() to support multiple collectors

TRebirthC:
Overload groupBy() and grouping() to support multiple collectors


lukaseder:
Thanks for your suggestion. That's a lot of overloads. It hasn't been decided yet, whether we want to go ahead with #154.

Issue #226, opened 2016-03-22
Add Function[N].compose() and .andThen() methods


lukaseder:
These should work along the lines of the methods in `java.util.function.Function`

patrox:
hi @lukaseder, could you please have a look at: https://github.com/jOOQ/jOOL/pull/270 and tell me if this approach is correct ?

lukaseder:
@patrox: Thanks for your suggestion. Your implementation suggestion is along the lines of what I must've had in mind at the time. However, I currently don't see the point of this anymore. A `Function[N]` always returns a single value, so it can only ever be composed with `Function1`... I guess the feature request is not that useful...?

lukaseder:
I'm closing this feature request for now, as I'm not convinced the functionality is useful.

PR #374, opened 2020-05-31
[#226] Add Function[N].compose and Function[N].andThen

kbb00:
.andThen() can be composed with all Function[N] and .compose() can be overloaded in Function1


lukaseder:
Thank you very much for your suggestion. Please note that just because there's a feature request on the roadmap doesn't mean we're going to implement it. In fact, these methods aren't that useful with `Function[N]` types, because they can only ever be used to compose a `Function[N]` and a `Function1`.

I'm not convinced we need these.

Issue #360, opened 2019-09-26
Add Agg.stddev() and variance() and similar aggregate functions


lukaseder:
It would be useful to be able to calculate some statistics functions like `stddev`, `variance` and other things, including linear regression functions.

Lu-Jichen:
I would like to work on this issue. I think to make the function more clear and easy to use, a mapping function which can map the object to a double value is needed. The return value should also be a double value, which is the standard deviation or the variance.

lukaseder:
Integrated with some minor improvements. Thanks again for your efforts!

PR #375, opened 2020-05-31
[#360] Add Agg.stddev() and variance()

kbb00:
Add Agg.stddev() and variance(), as well as a method Agg.linearRegression() which is used to do linear regression.

### DIFF
--- a/jOOL/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL/src/main/java/org/jooq/lambda/Agg.java
@@ -926,4 +926,234 @@ public class Agg {
             s -> s.map(Objects::toString).orElse("")
         );
     }
+	
+	public static <T extends Number> Collector<T, ?, Optional<Double>> stdDev() {
+
+        return Collector.of(
+                (Supplier<ArrayList<Double>>) ArrayList::new,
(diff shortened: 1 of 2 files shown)


lukaseder:
Thanks for your suggestion. I will comment your commit directly.

jOOL/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -926,4 +926,234 @@ else if (percentile == 1.0)
...
+
+                    for (final Tuple2<Double, Double> element : l) {
+                        xTimesy = xTimesy + (element.v1 - xAverage) * (element.v2 - yAverage);
+                    }
+                    final double bValue = xTimesy / variance;
+                    final double aValue = yAverage - bValue * xAverage;
+                    return Optional.of(tuple(bValue, aValue));
+                }
+        );
+    }
```
lukaseder, line 1158: SQL knows quite a few linear regression functions. Which one is this?

lukaseder:
I'm closing this PR for now, as I don't agree with the implementations. If you're interested in taking another go at this, please open a new PR.

Issue #376, opened 2020-06-03
Implement shortcut for 0.0 and 1.0 percentile calculations


lukaseder:
Th `0.0` percentile corresponds to `min()` and the `1.0` percentile corresponds to `max()`. We could implement this shortcut to improve the performance of these calculations, which can be performed in `O(N)` (without a sort), rather than `O(N log N)`

----
See also: https://github.com/jOOQ/jOOL/pull/372

lukaseder:
This improvement breaks the following test:

```java
assertEquals(Optional.of("b"), Stream.of("a", "b").collect(percentileBy(1.0, String::length)));
```

It now returns `"a"` instead of `"b"`. While this may be an unspecified implementation detail, users might have come to rely on the status quo. Let's be careful here.

JingHuaMan:
> It now returns "a" instead of "b".

Modifying the comparator can solve this problem: If two items are equal in value, force the output of the comparator to be minus. Then items that appear earlier in the input are more likely to appear earlier in the seq.

lukaseder:
> Modifying the comparator can solve this problem: If two items are equal in value, force the output of the comparator to be minus. Then items that appear earlier in the input are more likely to appear earlier in the seq.

As commented in the PR, this is an invalid technique. A `Comparator` contract mandates that if `a < b`, then `b > a`. Your implementation violates this for `a == b` cases, where a random operand is smaller than the other based on some observed fact (notably that this seems to have worked for our tests, but depending on the sort algorithm, it might as well not work).

PR #381, opened 2021-04-18
Feature issue 376

JingHuaMan:
https://github.com/jOOQ/jOOL/issues/376

If the `percentile` is 0 or 1, it's unnecessary to sort all the elements in the container before getting the output by the index; instead, we can just scan all the items in one traversal and get the minimum or maximum with a time complexity of O(n).

All the tests passed after the modification. Besides, in the test `testPercentileWithStringsAndFunction`, since the function is `String::length` and each string in the stream is of length 1, the sorting result should be the same as the original order of the input elements. 

I think the orginal test is not enough, so I add a new test for the function `percentileBy`.

Edit: add a second test.

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
@@ -911,6 +911,17 @@ public class Agg {
         if (percentile < 0.0 || percentile > 1.0)
             throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");
 
+        if (percentile == 0.0)
+            // If percentile is 0, this is the same as taking the item with the minimum value.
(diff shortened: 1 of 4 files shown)


lukaseder:
Thanks a lot for this suggestion. I agree with it. Will merge for inclusion in 0.9.15

lukaseder:
It seems I can't merge it anymore since you closed it. Is there any reason for your closing it? If I merge your PR, you'll get credit for it in the commit history. Otherwise, I'll go ahead and copy the contents...

lukaseder:
Nevermind. It seems you've submitted another PR later: https://github.com/jOOQ/jOOL/pull/382

lukaseder:
Duplicate of https://github.com/jOOQ/jOOL/pull/382

PR #382, opened 2021-04-23
Implement shortcut for 0.0 and 1.0 percentile calculations, and add two new tests

JingHuaMan:
https://github.com/jOOQ/jOOL/issues/376

If the `percentile` is 0 or 1, it's unnecessary to sort all the elements in the container before getting the output by the index; instead, we can just scan all the items in one traversal and get the minimum or maximum with a time complexity of O(n).

All the tests passed after the modification. Besides, in the test `testPercentileWithStringsAndFunction`, since the function is `String::length` and each string in the stream is of length 1, the sorting result should be the same as the original order of the input elements. 

I think the orginal test is not enough, so I add two new tests for the function `percentileBy`.

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
@@ -905,7 +905,11 @@ public class Agg {
     }
 
     /**
-     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering or the ordering of the index.
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering.
+     *
+     * @param function map the items in the streams into values
+     * @param comparator comparator used for sorting the items
+     * @return a collector that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function
      */
     public static <T, U> Collector<T, ?, Optional<T>> percentileBy(double percentile, Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
         if (percentile < 0.0 || percentile > 1.0)
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
@@ -911,6 +911,18 @@ public class Agg {
         if (percentile < 0.0 || percentile > 1.0)
             throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");
 
+        // CS304 Issue link: https://github.com/jOOQ/jOOL/issues/376
+        if (percentile == 0.0)
+            // If percentile is 0, this is the same as taking the item with the minimum value.
+            return minBy(function, comparator);
+        else if (percentile == 1.0)
+            // If percentile is 1, this is the same as taking the item with the maximum value,
+            // If there are multiple maxima, take the last one.
+            return maxBy(function, (o1, o2) -> {
+                int compareResult = comparator.compare(o1, o2);
+                return compareResult == 0 ? -1 : compareResult;
+            });
+
         // At a later stage, we'll optimise this implementation in case that function is the identity function
         return Collector.of(
             () -> new ArrayList<Tuple2<T, U>>(),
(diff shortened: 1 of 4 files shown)


jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -906,11 +906,27 @@ else if (l1[0] == null)
 
     /**
      * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering.
+     *
+     * @param function map the items in the streams into values
+     * @param comparator comparator used for sorting the items
+     * @return a collector that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function
```
lukaseder, line 912: It's probably a good idea to document these in general, but I prefer this be done in a separate task, for the entire API, not just this method. I've created a new issue to track this: https://github.com/jOOQ/jOOL/issues/388.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -906,11 +906,27 @@ else if (l1[0] == null)
...
      */
     public static <T, U> Collector<T, ?, Optional<T>> percentileBy(double percentile, Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
         if (percentile < 0.0 || percentile > 1.0)
             throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");
 
+        // CS304 Issue link: https://github.com/jOOQ/jOOL/issues/376
+        if (percentile == 0.0)
+            // If percentile is 0, this is the same as taking the item with the minimum value.
+            return minBy(function, comparator);
```
lukaseder, line 921: The comment says the same thing as the method call, so it isn't really necessary. I'll remove it again after merging.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -906,11 +906,27 @@ else if (l1[0] == null)
...
+     *
+     * @param function map the items in the streams into values
+     * @param comparator comparator used for sorting the items
+     * @return a collector that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function
      */
     public static <T, U> Collector<T, ?, Optional<T>> percentileBy(double percentile, Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
         if (percentile < 0.0 || percentile > 1.0)
             throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");
 
+        // CS304 Issue link: https://github.com/jOOQ/jOOL/issues/376
```
lukaseder, line 918: I don't know what CS304 means. Some reference to some external tracking system? There's no need for this, I will remove it. The convention to track github issues (if necessary) would be to use:

```java
// [#376] Rationale
```

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -906,11 +906,27 @@ else if (l1[0] == null)
...
+        if (percentile == 0.0)
+            // If percentile is 0, this is the same as taking the item with the minimum value.
+            return minBy(function, comparator);
+        else if (percentile == 1.0)
+            // If percentile is 1, this is the same as taking the item with the maximum value,
+            // If there are multiple maxima, take the last one.
+            return maxBy(function, (o1, o2) -> {
+                int compareResult = comparator.compare(o1, o2);
+                return compareResult == 0 ? -1 : compareResult;
```
lukaseder, line 927: This hack violates the `Comparator` contract. We can't implement it like this.

lukaseder: This seems to be the correct way to implement this:

```java
collectingAndThen(maxAllBy(function, comparator), s -> s.findLast())
```

JingHuaMan: Emmm, I think my solution is not incorrect, because the function `maxBy` will not sort the elements but compare them one by one, and the input comparator will not be modified. So even though this implementation violates the design philosophy of `Comparator`, it works and there seems no potential problem with it.

It's perfectly fine to implement this with `maxAllBy`, but in the worst case the space complexity will be O(n). That's my concern.

lukaseder: It's incorrect. If `comparator.compare(o1, o2) == 0`, then `comparator.compare(o2, o1) == 0`, yet you return `-1` in both cases. I don't want to spend the time now to find an edge case where this breaks sorting algorithms, but it should be easy to get an intuition about how this hack feels very wrong

> because the function `maxBy` will not sort the elements but compare them one by one

You shouldn't rely on such an implementation detail.

> So even though this implementation violates the design philosophy of `Comparator`, it works and there seems no potential problem with it.

Famous last words :)

> It's perfectly fine to implement this with `maxAllBy`, but in the worst case the space complexity will be O(n). That's my concern.

I'm open to other suggestions, but correctness always beats performance.

JingHuaMan: Yes, you are right. Thanks for your reply!

PR #383, opened 2021-04-25
Implement of issue 360

### LINKED ISSUE #360: Add Agg.stddev() and variance() and similar aggregate functions

Lu-Jichen:
Fixed #360 
Since the standard deviation and the variance have similar calculation process, there is an extracted method named "getVariance", which calculate the variance of the given data.

As for the test cases, there includes the empty Seq test, and the test with self defined class, and other basic test. All the tests have passed with no conflicts with other tests.

### DIFF
--- a/jOOL/pom.xml
+++ b/jOOL/pom.xml
@@ -229,7 +229,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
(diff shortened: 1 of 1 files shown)


lukaseder:
Duplicate of https://github.com/jOOQ/jOOL/pull/387

PR #384, opened 2021-04-25
Fix #352

### LINKED ISSUE #352: Don't swap Range bounds by default
Currently, when using the `Range` type, bounds are swapped by default, if they are not in order. This can be convenient occasionally, but incorrect in other cases. With the suggestion of supporting (partially) unbounded ranges (#350), swapping may become wrong.

We should introduce new API that allows for swapping explicitly (akin to SQL's `BETWEEN SYMMETRIC`) and stop swapping implicitly.

**This is an incompatible change. We might not actually implement it.**

awnofwheat:
I create a new constructer to create tuple2 so that we can choose whether to swap bounds by default.

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/tuple/Range.java
@@ -32,20 +32,61 @@ public class Range<T extends Comparable<T>> extends Tuple2<T, T> {
 
     private static final long serialVersionUID = 1L;
 
+    public static boolean swap;
+
+    //CS304 Issue link: https://github.com/jOOQ/jOOL/issues/352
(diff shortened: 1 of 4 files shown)


lukaseder:
Thanks for your suggestion.

While I appreciate the idea of being able to occasionally swap elements of a `Tuple2`, that functionality could (if at all) be implemented simply by a `Tuple2<T1, T2>.swapAttributes(): Tuple2<T2, T1>` method. For the record, I don't think this utility pulls its weight, so I wouldn't want an alternative PR in this case.

We can't retain a `swap` flag in the `Tuple2` type. A `Tuple2` doesn't care about the fact that its creator may have meant for the attributes to be swapped.

For the future, please discuss your ideas first before providing a PR.

PR #385, opened 2021-05-20
#221 Add percentileAll and medianAll

JingHuaMan:
#221 

Unlike `percentileBy` and `medianBy`, the function `percentileAllBy` and `medianAllBy` will retrieve all the values of the specific percentile.

The order of the output elements is the same as the order of those elements in the input `Seq`.

I add a lot of test cases for the two functions, and all the previous and new tests are passed.

Though there is already unsubmitted commits related to issue #221  in May 2020, it seems that there were some bugs with his codes.

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
@@ -941,8 +941,108 @@ public class Agg {
     }
 
     /**
-     * Get a {@link Collector} that calculates the common prefix of a set of strings.
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
      */
+    public static <T extends Comparable<? super T>> Collector<T, ?, Seq<T>> percentileAll(double percentile) {
+        return percentileAllBy(percentile, t -> t, naturalOrder());
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
+     */
+    public static <T> Collector<T, ?, Seq<T>> percentileAll(double percentile, Comparator<? super T> comparator) {
+        return percentileAllBy(percentile, t -> t, comparator);
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
+     */
+    public static <T, U extends Comparable<? super U>> Collector<T, ?, Seq<T>> percentileAllBy(double percentile, Function<? super T, ? extends U> function) {
+        return percentileAllBy(percentile, function, naturalOrder());
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
+     */
+    public static <T, U> Collector<T, ?, Seq<T>> percentileAllBy(double percentile, Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
+        if (percentile < 0.0 || percentile > 1.0)
+            throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");
+
+        if (percentile == 0.0)
+            // If percentile is 0, this is the same as taking the items with the minimum value.
+            return minAllBy(function, comparator);
+        else if (percentile == 1.0)
+            // If percentile is 1, this is the same as taking the items with the maximum value.
+            return maxAllBy(function, comparator);
+
+        return Collector.of(
+                () -> new ArrayList<Tuple2<? extends T, U>>(),
+                (l, v) -> l.add(tuple(v, function.apply(v))),
+                (l1, l2) -> {
+                    l1.addAll(l2);
+                    return l1;
+                },
+                l -> {
+                    int size = l.size();
+
+                    if (size == 0)
+                        return Seq.empty();
+                    else if (size == 1)
+                        return Seq.of(l.get(0).v1);
+
+                    l.sort(Comparator.comparing(t -> t.v2, comparator));
+
+                    int left = 0, right = (int) -Math.round(-(size * percentile + 0.5)) - 1, mid;
+
+                    List<T> result = new ArrayList<>();
+                    U value = l.get(right).v2;
+                    // Search the first value equal to the value at the position of PERCENTILE by binary search
+                    while (left < right) {
+                        mid = left + (right - left) / 2;
+                        if (comparator.compare(l.get(mid).v2, value) < 0)
+                            left = mid + 1;
+                        else
+                            right = mid;
+                    }
+
+                    for (left++; left < size && comparator.compare(l.get(left).v2, value) == 0; left++)
+                        result.add(l.get(left).v1);
+                    return Seq.seq(result);
+                }
+        );
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
+     */
+    public static <T extends Comparable<? super T>> Collector<T, ?, Seq<T>> medianAll() {
+        return medianAllBy(t -> t, naturalOrder());
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
+     */
+    public static <T> Collector<T, ?, Seq<T>> medianAll(Comparator<? super T> comparator) {
+        return medianAllBy(t -> t, comparator);
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>PERCENTILE_DISC(percentile)</code> function given a specific ordering, producing multiple results.
+     */
+    public static <T, U extends Comparable<? super U>> Collector<T, ?, Seq<T>> medianAllBy(Function<? super T, ? extends U> function) {
+        return medianAllBy(function, naturalOrder());
+    }
+
+    /**
+     * Get a {@link Collector} that calculates the derived <code>MEDIAN()</code> function given natural ordering, producing multiple results.
+     */
+    public static <T, U> Collector<T, ?, Seq<T>> medianAllBy(Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
+        return percentileAllBy(0.5, function, comparator);
+    }
+
+        /**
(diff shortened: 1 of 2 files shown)


jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -941,8 +941,118 @@ else if (percentile == 1.0)
     }
 
     /**
-     * Get a {@link Collector} that calculates the common prefix of a set of strings.
+     * CS304 Issue link: https://github.com/jOOQ/jOOL/issues/221
```
lukaseder, line 944: May I ask what CS304 means? I'll remove it again, but I'm curious. Is this some CS=Computer Science university class?

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -941,8 +941,118 @@ else if (percentile == 1.0)
...
+                        else
+                            right = mid;
+                    }
+
+                    for (; left < size && comparator.compare(l.get(left).v2, value) == 0; left++)
+                        result.add(l.get(left).v1);
+                    return Seq.seq(result);
+                }
+        );
+    }
```
lukaseder, line 1019: Have you checked, is there any refactoring potential to have common logic between `percentileBy()` and `percentileAllBy()`?

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -941,8 +941,118 @@ else if (percentile == 1.0)
...
+
+        if (percentile == 0.0)
+            // If percentile is 0, this is the same as taking the items with the minimum value.
+            return minAllBy(function, comparator);
+        else if (percentile == 1.0)
+            // If percentile is 1, this is the same as taking the items with the maximum value.
+            return maxAllBy(function, comparator);
+
+        return Collector.of(
+                () -> new ArrayList<Tuple2<? extends T, U>>(),
```
lukaseder, line 983: Why is this covariance needed here?

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: I mean, at least the collection to the accumulator seems to be the same, the difference being the finisher implementation.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -941,8 +941,118 @@ else if (percentile == 1.0)
...
+
+                    l.sort(Comparator.comparing(t -> t.v2, comparator));
+
+                    int left = 0;
+                    int right = (int) -Math.round(-(size * percentile + 0.5)) - 1;
+                    int mid;
+
+                    List<T> result = new ArrayList<>();
+                    U value = l.get(right).v2;
+                    // Search the first value equal to the value at the position of PERCENTILE by binary search
```
lukaseder, line 1005: Nice thinking!

lukaseder: Perhaps use `Collections.binarySearch()` instead, though?

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
JingHuaMan: I think we should apologize for these comments. We add these PRs because it is our software engineering course homework to make contributions on some open project╮(╯﹏╰）╭

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
JingHuaMan: I don't like the similar structure of `percentileBy` and `percentileAllBy` too, but I didn't find a proper way to refactor these codes.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
JingHuaMan: I didn't know this API before. It's obviously better to use Java standard library.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: No worries, just curious. I'll remove the comments. Another question, what made you choose jOOλ?

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: OK, I'll refactor it

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Sure, no worries. I'll refactor it. Thanks for the effort!

jOOL-java-8/src/test/java/org/jooq/lambda/CollectorTests.java
```diff
@@ -308,7 +297,225 @@ public void testPercentileWithStringsAndFunction() {
...
+
+            @Override
+            public int compareTo(Item o) {
+                return this.val - o.val;
+            }
+
+            Item reverse() {
+                return new Item(-this.val);
+            }
+        }
```
lukaseder, line 508: I'll merge this, but I'll probably get rid of this class. I'm not sure what value it provides compared to just using `Integer`

lukaseder:
Thanks again for your contribution. Note that if you plan on making more contributions, it will be sufficient to provide an implementation for the `jool` module only, the `jool-java-8` module is created automatically using scripts that aren't on this github repository.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: This should do the trick:

```java
private static <T, U, R> Collector<T, List<Tuple2<T, U>>, R> percentileCollector(
    Function<? super T, ? extends U> function,
    Comparator<? super U> comparator,
    Supplier<R> onEmpty,
    Function<T, R> onSingle,
    Function<List<Tuple2<T, U>>, R> finisher
) {
    return Collector.of(
        () -> new ArrayList<>(),
        (l, v) -> l.add(tuple(v, function.apply(v))),
        (l1, l2) -> {
            l1.addAll(l2);
            return l1;
        },
        l -> {
            int size = l.size();

            if (size == 0)
                return onEmpty.get();
            else if (size == 1)
                return onSingle.apply(l.get(0).v1);

            l.sort(Comparator.comparing(t -> t.v2, comparator));
            return finisher.apply(l);
        }
    );
}
```

Leading to e.g.

```java
public static <T, U> Collector<T, ?, Optional<T>> percentileBy(double percentile, Function<? super T, ? extends U> function, Comparator<? super U> comparator) {
    if (percentile < 0.0 || percentile > 1.0)
        throw new IllegalArgumentException("Percentile must be between 0.0 and 1.0");

    if (percentile == 0.0)
        return minBy(function, comparator);
    else if (percentile == 1.0)
        return collectingAndThen(maxAllBy(function, comparator), s -> s.findLast());
    else
        return percentileCollector(
            function,
            comparator,
            Optional::empty,
            Optional::of,
            l -> Optional.of(l.get(percentileIndex(percentile, l.size())).v1)
        );
}
```

Where 

```java
private static int percentileIndex(double percentile, int size) {

    // x.5 should be rounded down
    return (int) -Math.round(-(size * percentile + 0.5)) - 1;
}
```

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: The JDK method finds a random index if multiple values have the same index, so we can't use that.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -941,8 +941,118 @@ else if (percentile == 1.0)
...
+                        mid = left + (right - left) / 2;
+                        if (comparator.compare(l.get(mid).v2, value) < 0)
+                            left = mid + 1;
+                        else
+                            right = mid;
+                    }
+
+                    for (; left < size && comparator.compare(l.get(left).v2, value) == 0; left++)
+                        result.add(l.get(left).v1);
+                    return Seq.seq(result);
```
lukaseder, line 1016: The auxiliary `result` list is not necessary. We could use `l` again here and add the comparator predicate as a `Seq.filter()`. Not sure though if we want to keep a reference to the possibly large array list, though... Edge cases...

jOOL-java-8/src/test/java/org/jooq/lambda/CollectorTests.java
lukaseder: Nevermind. I see the point now. I'll just move the local class out of the method for better reuse.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
JingHuaMan: Yes, `Collections.binarySearch` will not return the first element equal to the target item. Part of the source codes of `Collections.binarySearch` attached.

```
while (low <= high) {
      int mid = (low + high) >>> 1;
      Comparable<? super T> midVal = get(i, mid);
      int cmp = midVal.compareTo(key);

      if (cmp < 0)
          low = mid + 1;
      else if (cmp > 0)
            high = mid - 1;
      else
            return mid; // key found
}
```

lukaseder:
Additional changes here: https://github.com/jOOQ/jOOL/commit/d41bd549ebad07efc1010b841e037a493796c689

lukaseder:
And here: https://github.com/jOOQ/jOOL/commit/1bdd69b2fd6f24fe193fbb53076cad0f33123108

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
JingHuaMan: It's mainly because this project is easy to learn, and it seems to be an important project because it's popular. Actually we have also considered JOOQ, but it's too big.

lukaseder:
PS: Feel free to add yourself as `@author` to the `Agg` class

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Makes sense. Yes, it's much harder to make meaningful contributions to jOOQ - especially because the tests are not open source.

PR #386, opened 2021-05-21
implement of issue 151, linear regression

### LINKED ISSUE #151: Add support for linear regression functions as collectors
All of these would be useful:
https://docs.oracle.com/database/121/SQLRF/functions165.htm#SQLRF00696

Lu-Jichen:
Fixed #151 

All of the functions follows the link of https://docs.oracle.com/database/121/SQLRF/functions165.htm#SQLRF00696. The formula of calculations are standard formulas. There are totally 9 functions implemented, as mentioned in the link.

As for the test cases, there includes the empty Seq test, and the test with self defined class and Tuple2. All the tests have passed with no conflicts with other tests.

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
@@ -979,4 +979,694 @@ public class Agg {
             s -> s.map(Objects::toString).orElse("")
         );
     }
+
+    //CS304 Issue link: https://github.com/jOOQ/jOOL/issues/151
+
+    /**
+     * Calculate the linear regression function Slope of the given object collectors,
+     * based on the mapping function from object to double number.
+     *
+     * <p><pre> Usage of aggregation function regrSlopeBy():
+     * The mapping functions functionX and functionY are functions mapping the objects to double values X and Y, for instance:
+     * {@code
+     * FunctionX<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v1);
+     * FunctionY<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v2);
+     * }
+     * The specific usage of varianceBy is like:
+     * {@code
+     * Seq.of(new Tuple2<Integer,Integer>(1,1),new Tuple2<Integer,Integer>(2,2)).collect(Agg.regrSlopeBy(functionX,functionY));
+     * }
+     * Besides, self defined class is also allowed with mapping function:
+     * {@code
+     * Seq.of(new Node(-1,1),new Node(1,1)).collect(Agg.regrSlopeBy(functionX,functionY));
+     * }
+     * </pre>
+     *
+     * @param functionX mapping function from object to double value representing x
+     * @param functionY mapping function from object to double value representing y
+     * @return the Slope value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-05-17
+     */
+    public static <T> Collector<T, ?, Optional<Double>> regrSlopeBy(Function<? super T, ? extends Number> functionX, Function<? super T, ? extends Number> functionY) {
+        class Accumulator {
+            Integer n = 0;
+            Double sumXY = 0.0;
+            Double sumXX = 0.0;
+            Double sumX = 0.0;
+            Double sumY = 0.0;
+        }
+
+        return Collector.of(
+                () -> new Accumulator(),
+                (a, t) -> {
+                    Double x = (Double) functionX.apply(t);
+                    Double y = (Double) functionY.apply(t);
+                    if (x != null && y != null) {
+                        a.n++;
+                        a.sumXY += x * y;
+                        a.sumXX += x * x;
+                        a.sumX += x;
+                        a.sumY += y;
+                    }
+                },
+                (a1, a2) -> {
+                    a1.n += a2.n;
+                    a1.sumX += a2.sumX;
+                    a1.sumY += a2.sumY;
+                    a1.sumXX += a2.sumXX;
+                    a1.sumXY += a2.sumXY;
+                    return a1;
+                },
+                a -> {
+                    if (a.n == 0 || a.n == 1) {
+                        return Optional.empty();
+                    }
+                    if ((a.sumXX - a.sumX * a.sumX / a.n) == 0.0) {
+//                        if (a.sumX==0){
+//                            if (a.sumY==0){
+//                                return Optional.of(0.0);
+//                            }else {
+//                                return Optional.of(Double.MAX_VALUE);
+//                            }
+//                        }
+//                        return Optional.of(a.sumY/a.sumX);
+                        return Optional.of(0.0);
+                    }
+                    return Optional.of((a.sumXY - a.sumX * a.sumY / a.n) / (a.sumXX - a.sumX * a.sumX / a.n));
+
+                }
+        );
+    }
+
+
+    //CS304 Issue link: https://github.com/jOOQ/jOOL/issues/151
+
+    /**
+     * Calculate the linear regression function Intercept of the given object collectors,
+     * based on the mapping function from object to double number.
+     *
+     * <p><pre> Usage of aggregation function regrInterceptBy():
+     * The mapping functions functionX and functionY are functions mapping the objects to double values X and Y, for instance:
+     * {@code
+     * FunctionX<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v1);
+     * FunctionY<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v2);
+     * }
+     * The specific usage of varianceBy is like:
+     * {@code
+     * Seq.of(new Tuple2<Integer,Integer>(1,1),new Tuple2<Integer,Integer>(2,2)).collect(Agg.regrInterceptBy(functionX,functionY));
+     * }
+     * Besides, self defined class is also allowed with mapping function:
+     * {@code
+     * Seq.of(new Node(-1,1),new Node(1,1)).collect(Agg.regrInterceptBy(functionX,functionY));
+     * }
+     * </pre>
+     *
+     * @param functionX mapping function from object to double value representing x
+     * @param functionY mapping function from object to double value representing y
+     * @return the Intercept value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-05-17
+     */
+    public static <T> Collector<T, ?, Optional<Double>> regrInterceptBy(Function<? super T, ? extends Number> functionX, Function<? super T, ? extends Number> functionY) {
+        class Accumulator {
+            Integer n = 0;
+            Double sumXY = 0.0;
+            Double sumXX = 0.0;
+            Double sumX = 0.0;
+            Double sumY = 0.0;
+        }
+
+        return Collector.of(
+                () -> new Accumulator(),
+                (a, t) -> {
+                    Double x = (Double) functionX.apply(t);
+                    Double y = (Double) functionY.apply(t);
+                    if (x != null && y != null) {
+                        a.n++;
+                        a.sumXY += x * y;
+                        a.sumXX += x * x;
+                        a.sumX += x;
+                        a.sumY += y;
+                    }
+                },
+                (a1, a2) -> {
+                    a1.n += a2.n;
+                    a1.sumX += a2.sumX;
+                    a1.sumY += a2.sumY;
+                    a1.sumXX += a2.sumXX;
+                    a1.sumXY += a2.sumXY;
+                    return a1;
+                },
+                a -> {
+                    if (a.n == 0 || a.n == 1) {
+                        return Optional.empty();
+                    }
+                    if ((a.sumXX - a.sumX * a.sumX / a.n) == 0.0) {
+//                        return Optional.of(0.0);
+                        return Optional.of(a.sumY / a.n);
+                    }
+                    return Optional.of(a.sumY / a.n - (a.sumXY - a.sumX * a.sumY / a.n) / (a.sumXX - a.sumX * a.sumX / a.n) * a.sumX / a.n);
+
+                }
+        );
+    }
+
+
+    //CS304 Issue link: https://github.com/jOOQ/jOOL/issues/151
+
+    /**
+     * Calculate the linear regression function R2 of the given object collectors,
+     * based on the mapping function from object to double number.
+     *
+     * <p><pre> Usage of aggregation function regrR2By():
+     * The mapping functions functionX and functionY are functions mapping the objects to double values X and Y, for instance:
+     * {@code
+     * FunctionX<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v1);
+     * FunctionY<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v2);
+     * }
+     * The specific usage of varianceBy is like:
+     * {@code
+     * Seq.of(new Tuple2<Integer,Integer>(1,1),new Tuple2<Integer,Integer>(2,2)).collect(Agg.regrR2By(functionX,functionY));
+     * }
+     * Besides, self defined class is also allowed with mapping function:
+     * {@code
+     * Seq.of(new Node(-1,1),new Node(1,1)).collect(Agg.regrR2By(functionX,functionY));
+     * }
+     * </pre>
+     *
+     * @param functionX mapping function from object to double value representing x
+     * @param functionY mapping function from object to double value representing y
+     * @return the R2 value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-05-17
+     */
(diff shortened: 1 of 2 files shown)


jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+     * FunctionX<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v1);
+     * FunctionY<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v2);
+     * }
+     * The specific usage of varianceBy is like:
+     * {@code
+     * Seq.of(new Tuple2<Integer,Integer>(1,1),new Tuple2<Integer,Integer>(2,2)).collect(Agg.regrSlopeBy(functionX,functionY));
+     * }
+     * Besides, self defined class is also allowed with mapping function:
+     * {@code
+     * Seq.of(new Node(-1,1),new Node(1,1)).collect(Agg.regrSlopeBy(functionX,functionY));
```
lukaseder, line 1001: The `Node` class from the tests shouldn't be mentioned here, it is confusing readers. We have `Tuple2` types that are better suited (I think?)

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+     *
+     * <p><pre> Usage of aggregation function regrSlopeBy():
+     * The mapping functions functionX and functionY are functions mapping the objects to double values X and Y, for instance:
+     * {@code
+     * FunctionX<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v1);
+     * FunctionY<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v2);
+     * }
+     * The specific usage of varianceBy is like:
+     * {@code
+     * Seq.of(new Tuple2<Integer,Integer>(1,1),new Tuple2<Integer,Integer>(2,2)).collect(Agg.regrSlopeBy(functionX,functionY));
```
lukaseder, line 997: Better use `new Tuple2<>(1, 1)` or even `tuple(1, 1)` to simplify this.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+
+    /**
+     * Calculate the linear regression function Slope of the given object collectors,
+     * based on the mapping function from object to double number.
+     *
+     * <p><pre> Usage of aggregation function regrSlopeBy():
+     * The mapping functions functionX and functionY are functions mapping the objects to double values X and Y, for instance:
+     * {@code
+     * FunctionX<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v1);
+     * FunctionY<Tuple2<Integer,Integer>, Double> mapping = e -> Double.valueOf(e.v2);
```
lukaseder, line 993: We don't have such types `FunctionX` or `FunctionY`. Just call them `Function`

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+     * </pre>
+     *
+     * @param functionX mapping function from object to double value representing x
+     * @param functionY mapping function from object to double value representing y
+     * @return the Slope value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-05-17
+     */
+    public static <T> Collector<T, ?, Optional<Double>> regrSlopeBy(Function<? super T, ? extends Number> functionX, Function<? super T, ? extends Number> functionY) {
```
lukaseder, line 1012: Again, `by()` has a different meaning in jOOL. Look at `Agg.minBy(Function<T, U>)`. It finds the value `T` for which the function argument produces the minimum value `U`. It does not produce the minimum value `U`.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: If you want to clarify that these correspond to the `x, y` parameters of the linear regression function, you can still call the variables `mappingX` or just `x`

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Just like with the previous review of https://github.com/jOOQ/jOOL/pull/387, we don't want do force users to use `Double` only. We should be able to calculate linear regressions for all numeric data types, including `Float`, `Double`, `BigDecimal` (granted, statistical functions make a bit less sense for integer types, but I don't see why we should offer `Double` only)

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+                    }
+                    if ((a.sumXX - a.sumX * a.sumX / a.n) == 0.0) {
+//                        if (a.sumX==0){
+//                            if (a.sumY==0){
+//                                return Optional.of(0.0);
+//                            }else {
+//                                return Optional.of(Double.MAX_VALUE);
+//                            }
+//                        }
+//                        return Optional.of(a.sumY/a.sumX);
```
lukaseder, line 1054: Please remove comments

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+                    }
+                    if (sstot == 0.0) {
+                        return Optional.of(1.0);
+                    } else {
+                        return Optional.of(1 - ssreg / sstot);
+                    }
+
+                }
+        );
+    }
```
lukaseder, line 1239: I'm not going to review all the formulas as implemented here, but I'm positive that a lot more code reuse is possible. Linear regression functions tend to depend on one another, or on other statistical functions, such as `COUNT()`, `AVG()`, `VAR_POP()`, `COVAR_POP()`, `STDDEV_POP()`, etc. etc. (see formulas here: https://github.com/jOOQ/jOOQ/issues/11547).

I'd much rather there were many more composable building blocks improving:

- Code reuse
- Allowing for things to be calculated for any data type, with only a single implementation.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +979,694 @@ else if (percentile == 1.0)
...
+                },
+                (a1, a2) -> {
+                    a1.n += a2.n;
+                    return a1;
+                },
+                a -> {
+                    return Optional.of(Double.valueOf(a.n));
+                }
+        );
+    }
```
lukaseder, line 1293: For example (as mentioned before), this implementation could have certainly been done via a `Agg.filter(v -> v != null, Agg.count())` collector?

lukaseder:
Thanks for your suggestion. Unfortunately, there are way too many issues with the current PR, so I will reject it.

lukaseder:
The work is not lost, though. Let's walk through this, step by step. I'd love to have linear regression function and other statistical function support in jOOλ. I'll think about how to best proceed here.

I'll also have another look at your other suggestion.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
Lu-Jichen: I would like to use mappingX and mappingY, then.

Lu-Jichen:
Thank you for your reply, I will consider the code reuse of the implementation and other details you mentioned. I will try to pull request another version later.

lukaseder:
Thanks a lot. Looking forward! I'll update my comments with additoinal findings from merging the other issue.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Let's skip this for now. I think we don't really need much Javadoc in a first step, to speed things up... I'll come up with some more standardised Javadoc examples later.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Same, let's just remove all Javadoc for the time being.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Same, let's just remove all Javadoc for the time being.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: So, we'll ignore other data types at first and use primitive doubles for now. Each one of these collectors should have 2 versions:

```java
public static Collector<Double, ?, Optional<Double>> regrSlopeDouble() {
    return regrSlopeDouble(t -> t);
}

public static <T> Collector<T, ?, Optional<Double>> regrSlopeDouble(ToDoubleFunction<? super T> functionX, ToDoubleFunction<? super T> functionY) {
}
```

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: This is still my main concern here. There's *a lot* of repeated maths, which I find hard to review and maintain. Maybe it can be refactored?

lukaseder:
Maybe, I'll have a go at implementing `COVAR_POP` (population covariance) as an example... It can be used in `REGR_SLOPE()` and `REGR_SXY()` as well as in `REGR_R2()` indirectly.

lukaseder:
Here's the issue regarding `COVAR_POP`: https://github.com/jOOQ/jOOL/issues/389

lukaseder:
So, here's a really simple implementation of `COVAR_POP`:
https://github.com/jOOQ/jOOL/commit/f53247510b46425d1f4a71880a45c26fad3b8874

I've refactored the calculation of an average from a `List<T>` given a `ToDoubleFunction<? super T>`, which simplified also your previous `variance` implementations very much. I mean, calculating the variance is now only:

```java
    private static <T> double variance0(List<T> l, ToDoubleFunction<? super T> function) {
        double average = avg0(l, function);
        return avg0(l, t -> Math.pow(function.applyAsDouble(t) - average, 2));
    }
```

Lu-Jichen:
OK, I will try to add the code. I have already rewrite the `REGR_AVGX()`, `REGR_AVGY()` and `REGR_COUNT()`

lukaseder:
And here's an example for `CORR()`: https://github.com/jOOQ/jOOL/issues/390, https://github.com/jOOQ/jOOL/commit/11aebbd5014930146dbb54c435186ec5ce53e217

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -1163,4 +1163,180 @@ else if (l.size() == 1)
...
+    public static <T> Collector<T, ?, Optional<Long>> regrCountLong(ToDoubleFunction<? super T> functionX, ToDoubleFunction<? super T> functionY) {
+        return collectingAndThen(toList(), l -> {
+            if (l.isEmpty())
+                return Optional.empty();
+            long result = 0;
+            for (T t : l)
+                result += t==null ? 0:1;
+            return Optional.of(result);
+        });
+    }
```
lukaseder, line 1190: About these ones:

1. Why `Long` when the input and calculation is `Double` based?
2. What's the point of passing `functionX` and `functionY` when they're not used?
3. In SQL, `REGR_COUNT(x, y)` makes some sense because of the implied `NULL` semantics, though even there, it can be trivially emulated using `COUNT(x + y)`, so it is of limited use. In Java, the `null` semantics can be achieved equally trivially using `Collectors.filtering(v -> v != null, Collectors.counting())`, so I'm not convinced this is really useful
4. Given the above emulation, we shouldn't operate on a pre-collected list and then iterate over it again.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -1163,4 +1163,180 @@ else if (l.size() == 1)
...
+     */
+    public static <T> Collector<T, ?, Optional<Double>> regrAvgXDouble(ToDoubleFunction<? super T> functionX, ToDoubleFunction<? super T> functionY) {
+        return collectingAndThen(toList(), l -> {
+            if (l.isEmpty())
+                return Optional.empty();
+            else if (l.size() == 1)
+                return Optional.of(0.0);
+            return Optional.of(avg0(l,functionX));
+        });
+    }
```
lukaseder, line 1210: Again, this isn't really much different from the ordinary `Collectors.averagingDouble()` collector, so why do we need it? Just because SQL has it? I mean, we could offer it, but then it should just delegate to the relevant JDK `Collectors.averagingDouble()` call. But I doubt it's useful.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -1163,4 +1163,180 @@ else if (l.size() == 1)
...
+
+    /**
+     * Get a {@link Collector} that calculates the <code>regrAvgXDouble()</code> function.
+     */
+    public static <T> Collector<T, ?, Optional<Double>> regrAvgXDouble(ToDoubleFunction<? super T> functionX, ToDoubleFunction<? super T> functionY) {
+        return collectingAndThen(toList(), l -> {
+            if (l.isEmpty())
+                return Optional.empty();
+            else if (l.size() == 1)
+                return Optional.of(0.0);
```
lukaseder, line 1207: That's incorrect.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -1163,4 +1163,180 @@ else if (l.size() == 1)
...
+    /**
+     * Get a {@link Collector} that calculates the <code>regrSxxDouble()</code> function.
+     */
+    public static <T> Collector<T, ?, Optional<Double>> regrSxxDouble(ToDoubleFunction<? super T> functionX, ToDoubleFunction<? super T> functionY) {
+        return collectingAndThen(
+                collectors(covarianceDouble(functionX, functionX),
+                        count()),
+                t -> !t.v1.isPresent() || t.v2==0 ? Optional.empty() : Optional.of(t.v1.get() * t.v2 )
+        );
+    }
```
lukaseder, line 1248: At first, I thought that looks like `REGR_SXY()`, not `REGR_SXX()`, because of the `COVAR_POP()` usage, but then I realised, you passed `functionX` twice, which makes this effectively `varianceDouble()` (which was implemented recently by your colleague). Better use that for clarity.

lukaseder: But again (and here we go, getting PRs for ill-specified issues is what we get when we ill-specify issues): Is this really useful? The `REGR_SXX()`, `REGR_SXY()`, and `REGR_SYY()` functions are some weird auxiliary functions in SQL. This is just multiplying the count with variance of covariance.

But both variance and covariance are some formula divided by the count, so this cancels out. It would then be more efficient to have a collector that calculates `REGR_SXX()`, `REGR_SXY()`, and `REGR_SYY()`, and base the other functions on this.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Overall, I have some doubts about the utility of some of these, but it wasn't well specified, and people might expect SQL aggregate functions in this library without much change in semantics - that's fine. So, let's ignore bullet 2 for now.

Lu-Jichen:
I have updated the `AVG` related functions, but I don't clearly know what you commented on 'SXX' functions, should I write them by my self to get better performance?

lukaseder:
> but I don't clearly know what you commented on 'SXX' functions, should I write them by my self to get better performance?

No worries, I'm just thinking out loud. I'll have to review the formulas from here: https://github.com/jOOQ/jOOQ/issues/11547#issuecomment-846149074. I think there are more optimal ways to calculate these functions than using the original formulas from that issue, but I'll have to verify that first for both jOOQ and jOOλ. Let's focus on API design and correctness first.

lukaseder:
I'll merge and fix some issues on the status quo.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: Also, returning `Long` is probably the right thing to do here, thinking about it again. It's consistent with `Agg.count()`

lukaseder:
Fixes here: https://github.com/jOOQ/jOOL/commit/0052689b75416a150ff3842ca3d7b9b748e8f212. I'll review this once more next week, regarding:

- Proper `Optional` usage and `NULL` semantics
- More tests

PR #387, opened 2021-05-21
implementation of issue 360

### LINKED ISSUE #360: Add Agg.stddev() and variance() and similar aggregate functions

Lu-Jichen:
Fixed #360
Since the standard deviation and the variance have similar calculation process, there is an extracted method named "getVariance", which calculate the variance of the given data.

As for the test cases, there includes the empty Seq test, and the test with self defined class, and other basic test. All the tests have passed with no conflicts with other tests.

### DIFF
--- a/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
+++ b/jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
@@ -28,10 +28,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeSet;
-import java.util.function.Function;
-import java.util.function.Predicate;
-import java.util.function.ToIntFunction;
-import java.util.function.ToLongFunction;
+import java.util.function.*;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +976,149 @@ else if (percentile == 1.0)
...
+     * @param function mapping function from object to double value
+     * @param l        a list containing all the objects
+     * @param size     the size of the list in collector
+     * @return the variance value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-04-25
+     */
+    private static <T> double getVariance(
+            Function<? super T, Double> function,
```
lukaseder, line 1003: This should be using `ToDoubleFunction`

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +976,149 @@ else if (percentile == 1.0)
...
+     *
+     * @param function mapping function from object to double value
+     * @return the stddev value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-04-25
+     */
+    public static <T> Collector<T, ?, Optional<Double>> stddevBy(
+            Function<? super T,
+                    Double> function) {
```
lukaseder, line 1052: While I aprreciate that there can be an optimised implementation for `double` and `float`, I think this is missing the point of jOOL to be `Number` subtype agnostic in such cases. See e.g. `Agg.sum()`, which can sum any `Number` subtype with specialised implementations in `org.jooq.lambda.Sum`

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +976,149 @@ else if (percentile == 1.0)
...
+                },
+                l -> {
+                    int size = l.size();
+                    if (size == 0) {
+                        return Optional.empty();
+                    }
+                    double variance = getVariance(function, l, size);
+                    return Optional.of(variance);
+                }
+        );
```
lukaseder, line 1122: This entire collector is unnecessary. You can just use `Collectors.collectingAndThen(toList(), varianceCalculationHere)`

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +976,149 @@ else if (percentile == 1.0)
...
+     *
+     * @param function mapping function from object to double value
+     * @return the stddev value
+     * @version 1.0
+     * @author Jichen Lu
+     * @date 2021-04-25
+     */
+    public static <T> Collector<T, ?, Optional<Double>> varianceBy(
+            Function<? super T,
+                    Double> function) {
```
lukaseder, line 1106: `variance(Function<? super T, ? extends U>)` (not `varianceBy()` - the `by()` has a different meaning in jOOL) is certainly useful, but it should be based on `variance()`, the general purpose function that can calculate variance itself.

lukaseder:
I have to reject this PR, see various comments.

jOOL-java-8/src/test/java/org/jooq/lambda/CollectorTests.java
```diff
@@ -990,4 +992,74 @@ public void testDropping() {
...
+            }
+
+            public Double function() {
+                return (double) value;
+            }
+
+            public int getValue() {
+                return this.value;
+            }
+        }
```
lukaseder, line 1048: I don't understand the purpose of this `Node` class. Surely, there's an easier way to test the functionality without it.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
```diff
@@ -979,4 +976,149 @@ else if (percentile == 1.0)
...
+     * Function<Integer, Double> mapping = e -> Double.valueOf(e);
+     * }
+     * The specific usage of varianceBy is like:
+     * {@code
+     * Seq.of(1, 1, 1, 1).collect(Agg.varianceBy(mapping));
+     * }
+     * Besides, self defined class is also allowed with mapping function:
+     * {@code
+     * Seq.of(new Node(1), new Node(1), new Node(1), new Node(1)).collect(Agg.varianceBy(mapping1));
+     * }
```
lukaseder, line 1095: Why does the `Node` class from the tests leak into the Javadoc? Readers will not understand what it means.

lukaseder:
OK, let's do it like this. I'll merge this, fix all the issues I've mentioned and then we can discuss the results. If you're still willing to have another go at https://github.com/jOOQ/jOOL/pull/386, we can do that based on the improvements if you're interested.

jOOL-java-8/src/main/java/org/jooq/lambda/Agg.java
lukaseder: I've given this a bit more thought. I don't think that statistical functions really make sense with integer types, so we can skip those. Also, we have a precedent where we're offering only `Optional<Double>` as a collector result: `percentRank()`.

lukaseder: OK, an even better precedent. Following the `Collectors.summingInt()` etc. naming convention, we have a `Agg.bitOr()` collector that is generic, and a `Agg.bitOrInt()` that is primitive. That'll do. Your contribution will be stddevDouble

Issue #393, opened 2022-10-17
Add Gradle build with Error-Prone in addition to maven


magicprinc:
For those who don't use maven

Google's Error prone is a great thing to find potential problems.

My English writing abilities are quite limited.
Thus, sorry for short explanations, comments, and discussions. I have made several improvements required for an internal project and I just want to "give back to community".
There will be several following PRs.

lukaseder:
Thanks for your suggestion and the suggested PR.

I don't see why we should offer this. Please discuss suggestions first, then we *might* proceed to PRs, but often, they won't be necessary / rejected.

magicprinc:
As I already said: writing in English is a painful activity for me :-)
I have made all these changes for an internal projects.
So I will "PR" them all to you, and you can choose what you like.

PS: Thank You very much for this great project.

lukaseder:
> So I will "PR" them all to you, and you can choose what you like.

Please don't. That's very time consuming. I'll just reject them all and won't look at them. Thanks for your understanding.

Some more insights into the rationale behind this, here: https://github.com/jOOQ/jOOQ/blob/main/CONTRIBUTING.md

PR #394, opened 2022-10-17
[jOOQ/jOOL#393] Gradle build with Error-Prone and improved gitignores

magicprinc:


lukaseder:
Thanks for your suggestion. I'll comment here: https://github.com/jOOQ/jOOL/issues/393

Issue #398, opened 2022-10-19
Fix Javadoc errors, Error prone errors and most obvious warnings


magicprinc:
Follow-up of https://github.com/jOOQ/jOOL/issues/396

Java 8+ has very strict DocLint: generating Javadoc for jOOL fails with many errors.

Error-prone also breaks build with some errors and generates a lot of warnings (some of them are quite obvious).

As mentioned in [jOOQ/jOOL#396] I have fixed all errors and some warnings, so build is green (except one failed test from [#397]).

magicprinc:
1) I have tried hard to find the best "autocrlf" setting, but nevertheless I have probably missed it :-(
GitHub shows too many changes – this is because of different CrLfs. 
If you compare the files in IDEA, etc. – the situation is much less dramatic.

2) I hope, more than one commit in PR is ok :-(
I've had [eb3bc6c] already, but then I decided, as they said, to "go beyond"

lukaseder:
Thanks for your suggestion. I'll reject this issue and the PR. It mixes many concerns, making a review impossible.

> 1. I have tried hard to find the best "autocrlf" setting, but nevertheless I have probably missed it :-(
>    (or You've missed it first – because all changed files have "true" LF now ;-).

I'm fine with fixing this, but cleanly and thoroughly, not as a piggy backed huge change to something as simple as "a few javadoc errors"

Personally, I think it's simplest to just leave it as it is.

> 2\. I hope, more than one commit in PR is ok :-(

Yes, but more than one topic isn't. Impossible to review.

magicprinc:
"report simple bugs showing what you think should be fixed"

It would flood your list here with similar quite stupid issues!

I could propose a "more compact" solution:

1) "Replace in files"
`* @see {@link `
-with-
`* {@link `

2) Javadoc' DocLint wants tag CODE inside tag PRE, so
Replace in files:
`<code><pre>`
-with-
`<pre><code>`

`</pre></code>`
-with-
`</code></pre>`

`</pre><code>`
-with-
`</code></pre>`

magicprinc:
I have just tried these steps on a fresh cloned repo:

2 errors
670 warnings

After fixing these two errors, you will go to the next step: javadoc errors and warnings

magicprinc:
Two so called compile errors are

1) `jOOL\jOOL\src\main\java\org\jooq\lambda\Blocking.java:74: error: [ReturnValueIgnored] Return value of 'get' must be used
[...]

PR #399, opened 2022-10-20
#398 PR Fix Javadoc errors, Error prone errors and most obvious warnings

magicprinc:
https://github.com/jOOQ/jOOL/issues/398

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 target
 *.iml
+.gradle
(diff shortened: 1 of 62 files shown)


lukaseder:
PR rejected, rationale here: https://github.com/jOOQ/jOOL/issues/398#issuecomment-1286597852

lukaseder:
Thinking about this, maybe it's much simpler if you just report simple bugs showing what you think should be fixed, rather than sending PRs? I've never been a fan of PRs, personally... Issues are much simpler to work with. E.g. the excess type variables you've found? So easy to fix directly, rather than discussing a PR back and forth...

